PR #594, opened 2022-01-21
Add /version GET and POST

mat-if:
## Summary

This allows us to set a version and allow users to access the latest
provided version.

Won't be too hard to expand down the road if we want to surface version types for multiple things, but opted to not introduce all that extra complexity right now since we didn't have immediate plans for it. YAGNI and all.

## Testing Plan

Unit tests and tested manually with:
`http POST localhost:8003/version 'authorization:Bearer test' version==1.2.3`
`http localhost:8003/version`

## Breaking Change

```
[ ] Yes
```


src/version/version.service.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Injectable } from '@nestjs/common';
+import { PrismaService } from '../prisma/prisma.service';
+import { Version } from '.prisma/client';
+
+@Injectable()
+export class VersionService {
+  constructor(private readonly prisma: PrismaService) {}
+
+  async create(version: string): Promise<Version> {
+    return this.prisma.version.create({
```
rohanjadvani, line 14: Should we use a transaction?

src/version/version.service.ts
```diff
@@ -0,0 +1,28 @@
...
+      data: {
+        version,
+      },
+    });
+  }
+
+  async getLatest(): Promise<Version | null> {
+    return await this.prisma.version.findFirst({
```
rohanjadvani, line 22:
```suggestion
    return this.prisma.version.findFirst({
```

src/version/version.service.ts
```diff
@@ -0,0 +1,28 @@
...
+      orderBy: {
+        id: 'desc',
```
rohanjadvani, line 24: I doubt that we'll have this out of order, but what do you think about sorting by the version?

src/version/version.service.ts
```diff
@@ -0,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { Injectable } from '@nestjs/common';
+import { PrismaService } from '../prisma/prisma.service';
+import { Version } from '.prisma/client';
+
+@Injectable()
+export class VersionService {
```
rohanjadvani, line 10: Could we use plural in the file / class name for consistency?

```suggestion
export class VersionsService {
```

src/version/version.rest.module.ts
```diff
@@ -0,0 +1,15 @@
...
+import { VersionController } from './version.controller';
+import { VersionService } from './version.service';
+
+@Module({
+  controllers: [VersionController],
+  exports: [VersionService],
+  imports: [PrismaModule],
+  providers: [VersionService],
+})
+export class VersionRestModule {}
```
rohanjadvani, line 15:
```suggestion
export class VersionsRestModule {}
```

src/version/version.rest.module.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Module } from '@nestjs/common';
+import { PrismaModule } from '../prisma/prisma.module';
+import { VersionController } from './version.controller';
+import { VersionService } from './version.service';
+
+@Module({
+  controllers: [VersionController],
+  exports: [VersionService],
+  imports: [PrismaModule],
+  providers: [VersionService],
```
rohanjadvani, line 13: For consistency, what do you think about moving the Prisma related services to be within their own module (ex. `versions.module.ts`) and then leaving the controller here?

src/version/version.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ApiKeyGuard } from '../auth/guards/api-key.guard';
+import { VersionDto } from './dto/version.dto';
+import { VersionService } from './version.service';
+
+@ApiTags('Version')
+@Controller('version')
+export class VersionController {
+  constructor(private readonly versionService: VersionService) {}
+
+  @ApiOperation({ summary: 'Gets the version of the Iron Fish API' })
```
rohanjadvani, line 24: Will this be for the API? Or the `ironfish` package?

src/version/version.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+      transform: true,
+    }),
+  )
+  @Post()
+  async updateVersion(
+    @Query()
+    { version }: VersionDto,
+  ): Promise<string> {
+    const newVersion = await this.versionService.create(version);
+    return newVersion.version;
```
rohanjadvani, line 45: I think we can return the entire object here. We have been doing that for creation elsewhere.

src/version/version.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+@ApiTags('Version')
+@Controller('version')
+export class VersionController {
+  constructor(private readonly versionService: VersionService) {}
+
+  @ApiOperation({ summary: 'Gets the version of the Iron Fish API' })
+  @Get()
+  async version(): Promise<string> {
+    const latestVersion = await this.versionService.getLatest();
+    return latestVersion ? latestVersion.version : '';
```
rohanjadvani, line 28: I'll let you make the call here and don't have a strong opinion, but I prefer to return objects (even if it's a single field). That way we can add more fields easily (if we ever want to return the release date for example), deprecate old fields, etc.

src/version/version.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+  ValidationPipe,
+} from '@nestjs/common';
+import { ApiExcludeEndpoint, ApiOperation, ApiTags } from '@nestjs/swagger';
+import { ApiKeyGuard } from '../auth/guards/api-key.guard';
+import { VersionDto } from './dto/version.dto';
+import { VersionService } from './version.service';
+
+@ApiTags('Version')
+@Controller('version')
+export class VersionController {
```
rohanjadvani, line 21:
```suggestion
export class VersionsController {
```

src/version/version.service.spec.ts
```diff
@@ -0,0 +1,58 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
```
rohanjadvani, line 4: nit: Can remove new line

src/version/dto/version.dto.ts
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { IsSemVer } from 'class-validator';
+
+export class VersionDto {
```
rohanjadvani, line 6: Can we use `CreateVersionDto` for consistency?

prisma/migrations/20220121222436_add_versions/migration.sql
```diff
@@ -0,0 +1,8 @@
+-- CreateTable
+CREATE TABLE "versions" (
```
rohanjadvani, line 2: Traditionally we have been doing migrations in their own pull request and service / controller changes in separate  ones. You can if you'd like to separate it out, but this is pretty low-risk and won't be used so it's not necessary.

prisma/migrations/20220121222436_add_versions/migration.sql
```diff
@@ -0,0 +1,8 @@
+-- CreateTable
```
rohanjadvani, line 1: Can the migration be `create_versions` instead of `add_versions` for consistency?

src/version/version.service.ts
mat-if: I thought about that, but landed here because it allows for the ability to rollback / change version patterns without needing to refactor.

src/version/version.controller.ts
mat-if: Yeah, I think you're right. Originally I thought to keep it simple as possible, but returning a string is pretty non-standard and pigeonholes future updates quite a bit (or breaking changes). Good call.

prisma/migrations/20220121222436_add_versions/migration.sql
mat-if: Good call, https://github.com/iron-fish/ironfish-api/pull/596. Probably won't get this PR updated today, but it'll get there.

src/version/version.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+  UsePipes,
+  ValidationPipe,
+} from '@nestjs/common';
+import { ApiExcludeEndpoint, ApiOperation, ApiTags } from '@nestjs/swagger';
+import { ApiKeyGuard } from '../auth/guards/api-key.guard';
+import { VersionDto } from './dto/version.dto';
+import { VersionService } from './version.service';
+
+@ApiTags('Version')
+@Controller('version')
```
rohanjadvani, line 20: Can this be plural too for consistency?

src/versions/versions.controller.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  @ApiOperation({ summary: 'Gets the version of the Iron Fish package' })
+  @Get()
+  async version(): Promise<Record<string, SerializedVersion>> {
+    const latestVersion = await this.versionsService.getLatest();
+    if (latestVersion === null) {
+      return {};
+    }
+    return {
+      ironfish: serializedVersionFromRecord(latestVersion),
```
rohanjadvani, line 34: I like that we're returning an object now instead of the version string, but since I don't think we'll need the `created_at` field, what do you think about just returning the version directly?
```suggestion
      ironfish: latestVersion.version,
```

src/versions/versions.controller.ts
```diff
@@ -0,0 +1,55 @@
...
+      errorHttpStatusCode: HttpStatus.UNPROCESSABLE_ENTITY,
+      transform: true,
+    }),
+  )
+  @Post()
+  async updateVersion(
+    @Query()
+    { version }: CreateVersionDto,
+  ): Promise<SerializedVersion> {
+    return serializedVersionFromRecord(
```
rohanjadvani, line 51: Nice 👍

src/versions/versions.service.ts
```diff
@@ -0,0 +1,30 @@
...
+          version,
+        },
+      });
+    });
+  }
+
+  async getLatest(): Promise<Version | null> {
+    return this.prisma.version.findFirst({
+      orderBy: {
+        id: 'desc',
```
rohanjadvani, line 26: Your earlier comment regarding future version pattern changes makes sense for this sort. However, since the creation is restricted to semantic versioning currently, I think it'd be confusing for someone to look at this method for the first time to understand why it's working as intended. And although it's unlikely for requests to be sent out of order, this could break if we send incorrect / older versions.

The version sort seems cleaner and easier to understand. What do you think about using that, and if we need to support a different pattern, we can update the sort / clean up old records (since they'll be moot anyways) when the time comes?

src/versions/versions.controller.ts
mat-if: I'm leaning towards continuing to return an object. After you got me thinking about the shape of the return data, I felt like returning an object here makes it less of a breaking change down the line if we want to add/remove fields. Could be useful to see when a version was deployed, also. I'm not sure if there's an immediate use for it, but I could see that being useful to surface alongside the version when we start consuming this endpoint

src/versions/versions.service.ts
mat-if: Yeah, after thinking about this more, I'm okay with this, fix incoming. I also wanted to double check that we wouldn't have to implement sorting logic for it in the query, but postgres [seems smart enough](https://www.db-fiddle.com/f/nYSA56m1zvrUCtWbfnyTbz/0)

rohanjadvani (approved):

PR #621, opened 2022-02-01
Allow staging.testnet.ironfish.network in staging

NullSoldier:
## Summary
Adds staging.testnet.ironfish.network to cors so you can hit the staging API from there

## Testing Plan
Test on staging environment

## Breaking Change

```
[ ] Yes
```


rohanjadvani:
Going to update this environment variable in the staging API instead.

PR #692, opened 2022-03-11
Recreate event if deleted

leanthebean:
## Summary
A user had the wrong number of points for an entry. I deleted the entry and tried to add it again, but it didn't work. This PR is an attempt to fix that. 

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -356,7 +356,7 @@ export class EventsService {
     if (url) {
       metadata = { ...metadata, url: url };
       const existingEvent = await this.getEventByUrl(url);
-      if (existingEvent) {
+      if (existingEvent && existingEvent.deleted_at == null) {
         return {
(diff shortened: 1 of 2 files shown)


src/events/events.service.ts
```diff
@@ -356,7 +356,7 @@ export class EventsService {
     if (url) {
       metadata = { ...metadata, url: url };
       const existingEvent = await this.getEventByUrl(url);
-      if (existingEvent) {
+      if (existingEvent && existingEvent.deleted_at == null) {
```
rohanjadvani, line 359: Can this check be moved into `getEventByUrl`? For a given URL, if we have an event with a null `deleted_at` and one with a non-null `deleted_at`, the query can currently return the deleted event.

rohanjadvani (approved):

PR #698, opened 2022-03-14
fix(api): Fix pagination for getting previous chunks of users

rohanjadvani:
## Summary

Fix the previous page cursor in the ranking query to not return the top ranked users.

## Testing Plan

Manually tested

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/users/users.service.ts
+++ b/src/users/users.service.ts
@@ -277,8 +277,13 @@ export class UsersService {
           ELSE
             TRUE
         END
-      ORDER BY
-        rank ASC
+      ORDER BY 
+        CASE WHEN $2
+          THEN 
+            rank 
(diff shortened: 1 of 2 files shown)


src/users/users.service.ts
```diff
@@ -291,6 +296,12 @@ export class UsersService {
       countryCode,
       eventType,
     );
+    // If fetching a previous page, the ranks are sorted in opposite order.
+    // Reverse the data so the returned chunk is in ascending order.
+    if (before === undefined) {
```
brekk, line 301: Minor, but `if (!before) {` ?

brekk (approved):
Minor suggestion, but not a blocker.

PR #716, opened 2022-03-23
feat(api): Add anonymous IP submission to telemetry

rohanjadvani:
## Summary

Submit node addresses to telemetry without any association to any other node fields.

## Testing Plan

Manually verified submission in Influx.

## Breaking Change

```
[ ] Yes
[x] No
```


deekerno (approved):
Nice catch on `X-Forwarded-For`.

PR #749, opened 2022-04-08
Add NodeUptime table and migration

mat-if:
## Summary

Preparation for upcoming changes. We'll have a graphile job check this periodically and update to 0 when the threshold has been hit, telemetry endpoint will update this when it receives a request with the extra graffiti parameter

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -28,23 +28,35 @@ model Event {
 }
 
 model User {
-  id                  Int       @id @default(autoincrement())
-  created_at          DateTime  @default(now()) @db.Timestamp(6)
-  updated_at          DateTime  @default(now()) @updatedAt @db.Timestamp(6)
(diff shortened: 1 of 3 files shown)


prisma/schema.prisma
```diff
@@ -28,25 +28,37 @@ model Event {
...
   events              Event[]
+  node_uptime         NodeUptime?
 
   @@index([email], name: "index_users_on_email")
   @@index([graffiti], name: "index_users_on_graffiti")
   @@map("users")
 }
 
+model NodeUptime {
+  user            User     @relation(fields: [user_id], references: [id])
```
rohanjadvani, line 52: nit: I prefer to have separate `id` / independent primary keys unless it's a join table. I've found it to be cleaner and easier to manage, but either way

rohanjadvani (approved):

rohanjadvani (approved):

PR #751, opened 2022-04-11
Implement NodeUptime tracking on telemetry endpoint

mat-if:
## Summary

When the API receives a telemetry write request, and the graffiti is included in the request payload, we want to track that as a NodeUptime "check-in" in the DB. We only want to up date the check-in count once per hour.

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/node-uptimes/node-uptimes.service.spec.ts
+++ b/src/node-uptimes/node-uptimes.service.spec.ts
@@ -1,0 +1,129 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { INestApplication } from '@nestjs/common';
+import assert from 'assert';
+import faker from 'faker';
+import { v4 as uuid } from 'uuid';
+import { PrismaService } from '../prisma/prisma.service';
+import { bootstrapTestApp } from '../test/test-app';
+import { NodeUptimesService } from './node-uptimes.service';
+
+describe('NodeUptimesService', () => {
+  let app: INestApplication;
+  let nodeUptimesService: NodeUptimesService;
+  let prisma: PrismaService;
+
+  beforeAll(async () => {
+    app = await bootstrapTestApp();
+    nodeUptimesService = app.get(NodeUptimesService);
+    prisma = app.get(PrismaService);
+    await app.init();
+  });
+
+  afterAll(async () => {
+    await app.close();
+  });
+
+  const createUser = async () => {
+    return await prisma.user.create({
+      data: {
+        email: faker.internet.email(),
+        graffiti: uuid(),
+        country_code: faker.address.countryCode(),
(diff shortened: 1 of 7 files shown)


src/telemetry/telemetry.controller.ts
```diff
@@ -53,6 +59,14 @@ export class TelemetryController {
     }
 
     this.submitIpWithoutNodeFieldsToTelemetry(request);
+
+    if (graffiti) {
+      const user = await this.usersService.findByGraffiti(graffiti);
+      if (user == null) {
```
rohanjadvani, line 65:
```suggestion
      if (user === null) {
```

or

```ts
if (user) {
  ...
}
```

src/node-uptimes/node-uptimes.service.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    prisma = app.get(PrismaService);
+    await app.init();
+  });
+
+  afterAll(async () => {
+    await app.close();
+  });
+
+  const createUser = async () => {
+    return await prisma.user.create({
```
rohanjadvani, line 29:
```suggestion
    return prisma.user.create({
```

src/node-uptimes/node-uptimes.service.ts
```diff
@@ -0,0 +1,51 @@
...
+export class NodeUptimesService {
+  constructor(private readonly prisma: PrismaService) {}
+
+  async upsert(user: User): Promise<NodeUptime | null> {
+    const now = new Date();
+    const oneHourAgo = new Date();
+    oneHourAgo.setHours(now.getHours() - 1);
+
+    return this.prisma.$transaction(async (prisma) => {
+      const nodeUptime = await this.get(user);
```
rohanjadvani, line 20: Should we pass `prisma` into this?

rohanjadvani (approved):

src/node-uptimes/node-uptimes.service.ts
```diff
@@ -0,0 +1,59 @@
...
+          total_hours: {
+            increment: 1,
+          },
+        },
+        create: {
+          user_id: user.id,
+          last_checked_in: now.toISOString(),
+          total_hours: 0,
+        },
+      });
```
NullSoldier, line 41: ok well do the eventing piece afterwards too right?

NullSoldier (approved):

src/node-uptimes/node-uptimes.service.ts
NullSoldier: Do you normally not use dep injection to get services like this?

mat-if: For clarity, I wasn't using the transaction's `prisma` to do the `get`, I was using the one on the service which was likely not handling the transaction properly

src/node-uptimes/node-uptimes.service.ts
mat-if: Yeah, that'll be the next PR

PR #752, opened 2022-04-11
feat(api): Create `user_points`

rohanjadvani:
## Summary

Create table for user points

## Testing Plan

N/A

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "license": "MPL-2.0",
   "homepage": "https://github.com/iron-fish/ironfish-api",
   "engines": {
-    "node": "16.x"
+    "node": "17.x"
   },
   "keywords": [
     "api"
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "license": "MPL-2.0",
   "homepage": "https://github.com/iron-fish/ironfish-api",
   "engines": {
-    "node": "17.x"
+    "node": "16.x"
   },
   "keywords": [
     "api"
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   "license": "MPL-2.0",
   "homepage": "https://github.com/iron-fish/ironfish-api",
   "engines": {
-    "node": "16.x"
+    "node": "17.x"
```
mat-if, line 11: Curious why this version bump - 16 is an LTS release and the same version we use for the main codebase

NullSoldier: Yeah, `Production applications should only use Active LTS or Maintenance LTS releases.`

Also can we split this version bump out into another PR since it might be controversion?

rohanjadvani: Yeah. I'll split this out - was testing some stuff with Prisma and should have left it out.

prisma/migrations/20220411181259_create_user_points/migration.sql
```diff
@@ -0,0 +1,44 @@
...
+CREATE INDEX "index_user_points_on_community_contribution" ON "user_points"("community_contribution_points");
+
+-- CreateIndex
+CREATE INDEX "index_user_points_on_pull_request_merged" ON "user_points"("pull_request_merged_points");
+
+-- CreateIndex
+CREATE INDEX "index_user_points_on_social_media_promotion" ON "user_points"("social_media_promotion_points");
+
+-- AddForeignKey
+ALTER TABLE "user_points" ADD CONSTRAINT "user_points_user_id_fkey" FOREIGN KEY ("user_id") REFERENCES "users"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
```
NullSoldier, line 44: I'm surprised this isn't `ON DELETE CASCADE` for a one-to-one model, with `user_id` as the PK.

NullSoldier (approved):
I'm approving this but I don't think we should bump to non LTS node.

PR #755, opened 2022-04-12
feat(api): Set up indexer to refresh all user points

rohanjadvani:
## Summary

Create a utility Graphile Worker job which will refresh points and timestamps for all users in the `user_points` table.

## Testing Plan

Added unit tests

## Breaking Change

```
[ ] Yes
[x] No
```


NullSoldier (approved):
I tested this manually by nuking my user_points table and triggering `REFRESH_USERS_POINTS` and watching it get populated correctly.

PR #764, opened 2022-04-14
Add node uptime credit task

mat-if:
## Breaking Change

```
[ ] Yes
[ ] No
```


src/telemetry/telemetry.controller.ts
```diff
@@ -63,7 +67,13 @@ export class TelemetryController {
...
       const user = await this.usersService.findByGraffiti(graffiti);
       if (user) {
-        await this.nodeUptimes.upsert(user);
+        const uptime = await this.nodeUptimes.upsert(user);
+        if (uptime && uptime.total_hours >= NODE_UPTIME_CREDIT_HOURS) {
+          await this.graphileWorkerService.addJob(
+            GraphileWorkerPattern.CREATE_NODE_UPTIME_EVENT,
+            { userId: user.id },
+          );
+        }
```
NullSoldier, line 76: We might need to move this into the graphile job we added here in another PR, or else we may need a lock here?

NullSoldier, line 76: We might need to move this into the graphile job we added here in another PR, or else we may need a lock here?

NullSoldier:
Ok this just needs a rebase

NullSoldier (approved):

src/telemetry/telemetry.controller.ts
NullSoldier: ok we added a lock.

NullSoldier (approved):

PR #765, opened 2022-04-15
Fix tests not using .env.test

NullSoldier:
## Summary

Tests are still using your .env and putting test data into your
development database. I don't know how to fix it properly, but
test-app and `envFilePath` option is not working at all. Maybe
when we have more time we can debug it.

## Testing Plan
Run tests and check my dev database for more data.

## Breaking Change

```
[ ] Yes
[x] No
```


package.json
```diff
@@ -74,6 +74,7 @@
     "cluster": "0.7.7",
     "compression": "1.7.4",
     "dd-trace": "2.4.2",
+    "dotenv-cli": "5.1.0",
```
mat-if, line 77: Should this be a dev dependency?

mat-if (approved):

package.json
NullSoldier: opps, yes.

NullSoldier: Fixed.

PR #767, opened 2022-04-15
Remove no from breaking change in PR template

NullSoldier:
## Summary

This is an oppinionated change but I noticed the front end doesn't have
it, and I just have to check NO for every PR. It'll be nice not to have
to do that I think.

## Breaking Change

```
[ ] Yes
[x] No
```


deekerno (approved):
Screen Shot 2022-04-15 at 12 22 37 PM

The last `[x] No`.

PR #769, opened 2022-04-15
Quick fix to avoid tracking uptime prematurely

mat-if:
## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

NullSoldier:
ah yes, I was just wondering about this. Good change.

PR #770, opened 2022-04-18
Send back node uptime in user metrics

NullSoldier:
## Summary

They just weren't being sent back, but now they are. I also fixed tests to actually test for the 0 points because it should be 0.

## Testing Plan
Ran tests

## Breaking Change

```
[ ] Yes
```


brekk (approved):
Not sure why the tests are changing as they are, but this change seems fine

mat-if (approved):

NullSoldier:
The tests were not correct. They weren't testing the actual values returned, and it seemed wrong to test the values only for the new category.

PR #773, opened 2022-04-18
Change Leaderboard query to take event types

NullSoldier:
## Summary

The old leaderboard only took 1 single event type, now it takes multiple
event types which it then filters the leaderboard on. I'm not sure if this is
the only place this needs to be changed, but the user metrics was also
changed to reflect this.

## Testing Plan

TODO: Run and adjust leaderboard test.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/common/utils/enums.ts
+++ b/src/common/utils/enums.ts
@@ -17,5 +17,5 @@ export function isInStrEnum<T extends StrEnum<T>>(
   value: unknown,
   enumType: T,
 ): boolean {
-  return getEnumStrValues(enumType).filter((enumValue) => enumValue === value);
+  return !!getEnumStrValues(enumType).find((enumValue) => enumValue === value);
 }
--- a/src/common/utils/enums.ts
+++ b/src/common/utils/enums.ts
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+export type StrEnumValue<T> = T[keyof T];
+export type StrEnum<T> = Record<keyof T, string>;
+
+function getEnumStrValues<T extends StrEnum<T>>(
+  enumType: T,
+): Array<StrEnumValue<T>> {
+  return Object.values(enumType)
+    .filter((v) => typeof v === 'string')
+    .map((v) => v as StrEnumValue<T>);
+}
+
+export function isInStrEnum<T extends StrEnum<T>>(
+  value: unknown,
+  enumType: T,
+): boolean {
+  for (const enumValue of getEnumStrValues(enumType)) {
+    if (enumValue === value) {
+      return true;
+    }
+  }
+  return false;
+}
--- a/src/common/utils/enums.ts
+++ b/src/common/utils/enums.ts
@@ -17,10 +17,5 @@ export function isInStrEnum<T extends StrEnum<T>>(
   value: unknown,
   enumType: T,
 ): boolean {
-  for (const enumValue of getEnumStrValues(enumType)) {
-    if (enumValue === value) {
-      return true;
-    }
-  }
-  return false;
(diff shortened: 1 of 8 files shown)


src/common/utils/enums.ts
```diff
@@ -0,0 +1,26 @@
...
+  value: unknown,
+  enumType: T,
+): boolean {
+  for (const enumValue of getEnumStrValues(enumType)) {
+    if (enumValue === value) {
+      return true;
+    }
+  }
+  return false;
+}
```
brekk, line 26: Minor, but this could be a filter?

```js
function isInStrEnum(value, enumType) {
  return getEnumStrValues(enumType).filter(
    (x) => x === value)
  )
}
```

src/users/dto/users-query.dto.ts
```diff
@@ -32,8 +34,11 @@ export class UsersQueryDto extends PaginationArgsDto {
   @IsString()
   readonly search?: string;
 
-  @ApiPropertyOptional({ description: 'Event type filter', enum: EventType })
+  @ApiPropertyOptional({ description: 'Event types filter', enum: EventType })
   @IsOptional()
-  @IsEnum(EventType)
-  readonly event_type?: EventType;
+  @Transform(({ value }: TransformFnParams): unknown[] => {
+    return Array.isArray(value) ? (value as unknown[]) : [value];
```
brekk, line 40: I'd kinda hope that this kinda assertion / manipulation wasn't needed? Surprising to me that the type system doesn't account for this directly.

src/users/users.service.ts
```diff
@@ -334,12 +339,20 @@ export class UsersService {
       data[data.length - 1].rank,
       1,
       countryCode,
-      eventType,
+      !!eventTypes?.length,
```
brekk, line 342: Since this is used multiple places, storing a reference to it early in the function and re-using it makes sense?

```
const someEvents = !!eventTypes?.length
```

And then re-use it in 3 places?

brekk (approved):
Suggestions, no blockers -- I'd kinda hope that TS affords something other than hoops (per the second comment)

src/common/utils/enums.ts
NullSoldier: If you want it be a way you like, just turn it into a github suggestion and I'll click the button.

src/users/dto/users-query.dto.ts
NullSoldier: Do you mean casting to `unknown[]`? value is typed as `any` from 'class-validator' so you need to type it to match `unknown[]` as to not return any which will produce an eslint warning.

src/users/users.service.ts
NullSoldier: I'm guessing this  isbecause it's unclear what `!!eventTypes?.length` actually stands for, so I can factor it out to make that more clear into a descriptive variable.

NullSoldier: It should be more clear now.

src/common/utils/enums.ts
brekk, line 26:
```suggestion
export function isInStrEnum<T extends StrEnum<T>>(
  value: unknown,
  enumType: T,
): boolean {
  return getEnumStrValues(enumType).filter((enumValue) => enumValue === value);
}
```

src/users/dto/users-query.dto.ts
brekk: I guess it's just weird to me how much the type system isn't adding anything here.

src/common/utils/enums.ts
NullSoldier: Fixed

NullSoldier: I fixed the code, the suggestion used `filter()` but that needs to be `every()`

NullSoldier: Changed this to a find() call with boolean coercion from filter() since filter() returns an arrray of found values.

mat-if (approved):

PR #778, opened 2022-04-19
Add SEND_TRANSACTION event type

danield9tqh:
## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
Tested manually, I'll put up another PR to expose it in the metrics.

PR #779, opened 2022-04-20
Add SEND_TRANSACTION to user metrics API

NullSoldier:
## Summary
Exposing this now

## Testing Plan

Run tests and hit the API manually

## Breaking Change

```
[ ] Yes
```


brekk (approved):
:shipit:

PR #780, opened 2022-04-20
Add user filtering to leaderboard

NullSoldier:
## Summary

Added the ability to filter on user ids for the leaderboard so that
tests can probably test for rank calcalation. We still need to add a few
more tests for the leaderboard but this will make it easier now. No
previous tests actually checked the leaderboard logic until now.

## Testing Plan

Updated tests and ran them, I also checked production data locally.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/users/users.service.spec.ts
+++ b/src/users/users.service.spec.ts
@@ -215,24 +215,6 @@ describe('UsersService', () => {
   });
 
   describe('listWithRank', () => {
-    it('returns a chunk of users with rank', async () => {
-      const limit = 10;
-      const { data: records } = await usersService.listWithRank({
-        limit,
-        search: '7',
-      });
-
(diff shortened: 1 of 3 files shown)


src/users/users.service.spec.ts
```diff
@@ -215,24 +215,6 @@ describe('UsersService', () => {
...
-      expect(records.length).toBeLessThanOrEqual(limit);
-
-      for (const record of records) {
-        expect(record).toMatchObject({
-          id: expect.any(Number),
-          graffiti: expect.any(String),
-          rank: expect.any(Number),
-        });
-      }
-    });
```
NullSoldier, line 234: This test wasn't testing anything.

brekk (approved):
:shipit:

mat-if (approved):

PR #783, opened 2022-04-21
Add API endpoint to get ironfish deposit address for incentivized phase 2

danield9tqh:
## Summary
Returning deposit address for incentivized test net phase 2

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```


src/transactions/transactions.controller.spec.ts
```diff
@@ -387,4 +390,15 @@ describe('TransactionsController', () => {
       });
     });
   });
+
+  describe('GET /transactions/deposit_address', () => {
+    it('retuns deposit address from config', async () => {
+      const { body } = await request(app.getHttpServer())
+        .get('/transactions/deposit_address')
```
NullSoldier, line 397: Now that we have the deposit controller, since these are related to deposits, should we move this to `deposits/address`?

danield9tqh: yea I'm ok with that

NullSoldier (approved):

PR #790, opened 2022-04-26
Add Deposit Model

NullSoldier:
## Summary

Also add deposits/head() api to test that it works. A deposit
represents a single decrypted deposit from an end user that sent mony to
the "iron bank".

https://linear.app/ironfish/issue/IRO-1747/add-new-api-to-get-the-latest-head-to-process-transactions-for
https://linear.app/ironfish/issue/IRO-1796/add-new-deposit-model

## Testing Plan

Wrote tests and ran locally

## Breaking Change

```
[ ] Yes
```


prisma/schema.prisma
```diff
@@ -198,3 +198,16 @@ model UserPoints {
...
+  id                  Int                @id @default(autoincrement())
+  created_at          DateTime           @default(now()) @db.Timestamp(6)
+  updated_at          DateTime           @default(now()) @updatedAt @db.Timestamp(6)
+  transaction_hash    String             @db.VarChar
+  block_hash          String             @db.VarChar
+  block_sequence      Int
+  note_index          Int
+  network_version     Int
+  main                Boolean
+  amount              Int
```
rohanjadvani, line 212: nit: Can you add `@@map("deposits")` for consistency with other table names?

src/events/deposit.controller.spec.ts
```diff
@@ -0,0 +1,72 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { HttpStatus, INestApplication } from '@nestjs/common';
+import request from 'supertest';
+import { ApiConfigService } from '../api-config/api-config.service';
+import { PrismaService } from '../prisma/prisma.service';
+import { bootstrapTestApp } from '../test/test-app';
+
+describe('EventsController', () => {
```
rohanjadvani, line 10:
```suggestion
describe('DepositsController', () => {
```

src/events/deposit.controller.ts
```diff
@@ -0,0 +1,35 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Controller, Get, UseGuards } from '@nestjs/common';
+import { ApiOperation, ApiTags } from '@nestjs/swagger';
+import { ApiKeyGuard } from '../auth/guards/api-key.guard';
+import { DepositService } from './deposit.service';
+import { SerializedDeposit } from './dto/serialized-deposit.dto';
+
+@ApiTags('Deposit')
+@Controller('deposits')
+export class DepositController {
```
rohanjadvani, line 12: nit for consistency
```suggestion
export class DepositsController {
```

Can the file name be made plural too for consistency?

src/events/deposit.service.ts
```diff
@@ -0,0 +1,30 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Injectable } from '@nestjs/common';
+import { Deposit } from '@prisma/client';
+import { ApiConfigService } from '../api-config/api-config.service';
+import { SortOrder } from '../common/enums/sort-order';
+import { PrismaService } from '../prisma/prisma.service';
+
+@Injectable()
+export class DepositService {
```
rohanjadvani, line 11: Same nit as the controller
```suggestion
export class DepositsService {
```

src/events/dto/serialized-deposit.dto.ts
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+export interface SerializedDeposit {
```
rohanjadvani, line 5: We use DTOs for payloads that our controllers accept (POST bodies). Can this file be renamed and/or moved to the `interfaces` folder?

prisma/schema.prisma
NullSoldier: done

src/events/deposit.controller.spec.ts
NullSoldier: done

src/events/deposit.controller.ts
NullSoldier: done

src/events/deposit.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  constructor(
+    private readonly config: ApiConfigService,
+    private readonly prisma: PrismaService,
+  ) {}
+
+  async head(): Promise<Deposit | null> {
+    const networkVersion = this.config.get<number>('NETWORK_VERSION');
+
+    return await this.prisma.deposit.findFirst({
```
rohanjadvani:
```suggestion
    return this.prisma.deposit.findFirst({
```

src/events/deposit.controller.ts
NullSoldier: yeah I missed that, it was sloppy of me. Thanks.

src/events/deposit.service.ts
NullSoldier: done

src/events/dto/serialized-deposit.dto.ts
NullSoldier: That's confusing as DTO (data transfer object) indicates that it should be for transfer out, and transfer in. This class is used as the response to a request. Does that not mean it is a schema for transfering an object?

src/events/deposit.controller.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        },
+      });
+
+      response = await request(app.getHttpServer())
+        .get(`/deposits/head`)
+        .set('Authorization', `Bearer ${API_KEY}`)
+        .expect(HttpStatus.OK);
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      expect(response.body.deposit).toBeNull();
```
rohanjadvani: If you'd like to remove the eslint disabling:

```ts
expect(response.body).toMatchObject({
  deposit: null,
})
```

Unfortunately `supertest` doesn't have a way to use generics for the response, so we've been checking the response schema like this in other controllers.

src/events/deposit.controller.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        where: { id: deposit.id },
+      });
+
+      response = await request(app.getHttpServer())
+        .get(`/deposits/head`)
+        .set('Authorization', `Bearer ${API_KEY}`)
+        .expect(HttpStatus.OK);
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      expect(response.body.deposit.id).toEqual(deposit.id);
```
rohanjadvani: Same as above
```suggestion
expect(response.body).toMatchObject({
  deposit: {
    id: deposit.id
  },
})
```

prisma/migrations/20220426193548_create_deposit/migration.sql
```diff
@@ -0,0 +1,15 @@
+-- CreateTable
+CREATE TABLE "Deposit" (
```
rohanjadvani, line 2: Can you regenerate this migration file with the new mapped name?

src/events/dto/serialized-deposit.dto.ts
NullSoldier: I moved it, though I'm still not sure the distinction. It seems like both `interfaces` and `dtos` contain DTO's, but DTO's are purely request input validators.

src/events/deposit.controller.spec.ts
NullSoldier: We already disable other lint rules letting us use any in tests, what do you think about this one too? https://github.com/iron-fish/ironfish-api/pull/791

prisma/migrations/20220426193548_create_deposit/migration.sql
NullSoldier: done

src/events/dto/serialized-deposit.dto.ts
rohanjadvani: Yeah, the current pattern is using DTOs for incoming request data objects that we can run validation on using `class-validator`. I'm open to moving the serializable return types (which omit fields from our models) if that seems more consistent with the DTO name, but I've usually seen them used for validation within NestJS.

src/events/deposit.controller.spec.ts
NullSoldier: See comment here, https://github.com/iron-fish/ironfish-api/pull/790#discussion_r859139319

src/events/deposit.service.ts
NullSoldier: done

src/events/deposit.controller.spec.ts
rohanjadvani: Sure, that's fine with me. I think this is pretty safe since it's for tests, and contributors can use whatever they find has less friction.

NullSoldier: Removed the lint ignores since 791 merged.

src/events/deposit.controller.spec.ts
NullSoldier: Removed the lint ignores since https://github.com/iron-fish/ironfish-api/pull/791 merged.

src/events/deposits.controller.ts
```diff
@@ -0,0 +1,35 @@
...
+  async head(): Promise<{ deposit: SerializedDeposit | null }> {
+    const deposit = await this.depositService.head();
+
+    if (!deposit) {
+      return { deposit: null };
+    }
+
+    return {
+      deposit: {
+        id: deposit.id,
```
rohanjadvani, line 27: Not a blocker, but is there any preference between putting all these fields top-level versus under a `deposit` field?

NullSoldier: There is a bug in the serializer in that it turns a null response into an empty object, but only if it's the root return valule.

I'm fine being consistent with the other APIs, though I usually always put data in a key because if you ever need to add another top level object to return you're screwed.

NullSoldier: Fixed this by using 404 instead of returning null, and removing the nested key.

rohanjadvani (approved):

PR #792, opened 2022-04-26
Add Deposit Uploading

NullSoldier:
## Summary
For Phase 2 the Ironfish node will send transaction data that was sent to the the deposit address to the API. The API needs to create events for users if they deposited funds correctly. 

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -16,7 +16,7 @@ export const NODE_UPTIME_CREDIT_HOURS = 12; // Credit in 12 hour chunks
 
 // Send Transaction
 export const ORE_TO_IRON = 100000000;
-export const SEND_TRANSACTION_LIMIT = ORE_TO_IRON * 0.1; // Only transactions >= to this amount will get points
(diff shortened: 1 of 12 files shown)


src/common/constants.ts
```diff
@@ -14,6 +14,10 @@ export const MAX_LIMIT = 100;
 export const NODE_UPTIME_CHECKIN_HOURS = 1; // Check-in can happen once per hour
 export const NODE_UPTIME_CREDIT_HOURS = 12; // Credit in 12 hour chunks
 
+// Send Transaction
+export const ORE_TO_IRON = 100000000;
+export const SEND_TRANSACTION_LIMIT = ORE_TO_IRON * 0.1; // Only transactions >= to this amount will get points
```
NullSoldier, line 19: I think we should use the ORE suffix here, just like how we have _HOURS suffix on time to indicate the time unit

NullSoldier, line 19: I think we should use the ORE suffix here, just like how we have _HOURS suffix on time to indicate the time unit

danield9tqh (approved):

PR #795, opened 2022-04-27
test(api): Add minimum to mocked size in blocks controller

rohanjadvani:
## Summary

This test sometimes updates snapshots and throws a different validation error since the API validates for a positive size.

## Testing Plan

Ran unit tests.

## Breaking Change

```
[ ] Yes
```


deekerno (approved):

rohanjadvani:
I'm going to close and re-open this since Linear messed up ticket IDs so this is attached to the incorrect task.

PR #802, opened 2022-05-02
Update user points on event deletion

NullSoldier:
## Summary

When events are deleted, the UserPoints aggregate model is not
deleted. This causes the aggregate table to now be updated when an event
is deleted.

## Testing Plan
I wrote a test

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/user-points/user-points.service.ts
+++ b/src/user-points/user-points.service.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { Injectable } from '@nestjs/common';
+import { Injectable, NotFoundException } from '@nestjs/common';
 import { PrismaService } from '../prisma/prisma.service';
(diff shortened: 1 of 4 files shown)


src/user-points/user-points.service.ts
```diff
@@ -11,6 +11,20 @@ import { EventType, Prisma, UserPoints } from '.prisma/client';
 export class UserPointsService {
   constructor(private readonly prisma: PrismaService) {}
 
+  async findOrThrow(userId: number): Promise<UserPoints> {
```
rohanjadvani, line 14: Can we add a test, similar to the [`UsersService`](https://github.com/iron-fish/ironfish-api/blob/master/src/users/users.service.spec.ts#L61-L84)?

src/user-points/user-points.service.ts
```diff
@@ -70,13 +84,15 @@ export class UserPointsService {
         user_id: userId,
       },
     });
+
```
danield9tqh, line 87: 💯

danield9tqh (approved):

src/user-points/user-points.service.ts
NullSoldier: done

PR #806, opened 2022-05-04
Track deposits head

danield9tqh:
## Summary
We need to track blocks that do not have deposits in them as well as deposits. This is so the transaction sync process can restart from a block that does not have deposits in it.

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -222,3 +222,10 @@ model Deposit {
   @@map("deposits")
   @@unique([transaction_hash, graffiti], name: "uq_deposits_on_transaction_hash_and_graffiti", map: "uq_deposits_on_transaction_hash_and_graffiti")
 }
+
+model DepositHead {
+    id                Int                @id
+    block_hash        String             @db.VarChar
+
+    @@map("deposit_head")
+}
--- a/src/events/deposits.controller.spec.ts
+++ b/src/events/deposits.controller.spec.ts
@@ -5,30 +5,37 @@ import { HttpStatus, INestApplication } from '@nestjs/common';
 import { EventType, User } from '@prisma/client';
 import faker from 'faker';
 import request from 'supertest';
+import { v4 as uuid } from 'uuid';
 import { ApiConfigService } from '../api-config/api-config.service';
 import { BlockOperation } from '../blocks/enums/block-operation';
 import { ORE_TO_IRON } from '../common/constants';
 import { PrismaService } from '../prisma/prisma.service';
 import { bootstrapTestApp } from '../test/test-app';
 import { UsersService } from '../users/users.service';
-import { DepositsService } from './deposits.service';
-import { UpsertDepositsDto } from './dto/upsert-deposit.dto';
+import {
+  DepositTransactionDto,
+  UpsertDepositsDto,
+  UpsertDepositsNoteDto,
+  UpsertDepositsOperationDto,
+} from './dto/upsert-deposit.dto';
 
 describe('DepositsController', () => {
   let app: INestApplication;
   let config: ApiConfigService;
   let prisma: PrismaService;
-  let deposits: DepositsService;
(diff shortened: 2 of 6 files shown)


prisma/schema.prisma
```diff
@@ -222,3 +222,11 @@ model Deposit {
   @@map("deposits")
   @@unique([transaction_hash, graffiti], name: "uq_deposits_on_transaction_hash_and_graffiti", map: "uq_deposits_on_transaction_hash_and_graffiti")
 }
+
+model DepositHead {
+    id                Int                @id
+    block_hash        String             @db.VarChar
+    block_sequence    Int
+
+    @@map("deposithead")
```
rohanjadvani, line 231:
```suggestion
    @@map("deposit_head")
```

src/events/deposits.controller.spec.ts
```diff
@@ -244,4 +215,34 @@ describe('DepositsController', () => {
...
+  const notes = (amounts: number[], graffiti: string) => {
+    return amounts.map((amount) => {
+      return { memo: graffiti, amount: amount * ORE_TO_IRON };
+    });
+  };
+
+  const transaction = (notes: UpsertDepositsNoteDto[], hash?: string) => {
+    return {
+      hash: hash || uuid(),
+      notes: notes,
```
rohanjadvani, line 228:
```suggestion
      notes,
```

src/events/deposits.controller.spec.ts
```diff
@@ -244,4 +215,34 @@ describe('DepositsController', () => {
...
+  };
+
+  const depositOperation = (
+    transactions: DepositTransactionDto[],
+    type: BlockOperation,
+    hash?: string,
+    sequence?: number,
+  ): UpsertDepositsOperationDto => {
+    return {
+      type: type,
```
rohanjadvani, line 239:
```suggestion
      type,
```

src/events/deposits.controller.spec.ts
```diff
@@ -244,4 +215,34 @@ describe('DepositsController', () => {
...
+      block: {
+        hash: hash || uuid(),
+        timestamp: new Date(),
+        sequence: sequence || 0,
+      },
+      transactions: transactions,
```
rohanjadvani, line 245:
```suggestion
      transactions,
```

prisma/schema.prisma
```diff
@@ -222,3 +222,11 @@ model Deposit {
   @@map("deposits")
   @@unique([transaction_hash, graffiti], name: "uq_deposits_on_transaction_hash_and_graffiti", map: "uq_deposits_on_transaction_hash_and_graffiti")
 }
+
+model DepositHead {
+    id                Int                @id
```
NullSoldier, line 227: do we need this?

src/events/deposits.controller.ts
```diff
@@ -29,19 +29,16 @@ export class DepositsController {
...
+    const depositHead = await this.deposits.head();
 
-    if (!deposit) {
+    if (!depositHead) {
       throw new NotFoundException();
     }
 
     return {
-      id: deposit.id,
-      transaction_hash: deposit.transaction_hash,
```
rohanjadvani, line 41: Is this no longer needed?

prisma/schema.prisma
danield9tqh: yes so we know to update only the single head object with a constant id (1)

src/events/deposits.controller.ts
NullSoldier: it was never used, actually.

NullSoldier (approved):

prisma/schema.prisma
NullSoldier: I think this comment was on the old version, but it was already done on the non WIP.

PR #811, opened 2022-05-05
feat(api): Pull `total_points` from `user_points`

rohanjadvani:
## Summary

We can remove `users.total_points` since we have materialized values in `user_points`. Before removing that column, we should migrate existing reads to use the new table.

## Testing Plan

Updated unit tests.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/users/users.controller.ts
+++ b/src/users/users.controller.ts
@@ -32,6 +32,7 @@ import { PaginatedList } from '../common/interfaces/paginated-list';
 import { EventsService } from '../events/events.service';
 import { SerializedEventMetrics } from '../events/interfaces/serialized-event-metrics';
 import { NodeUptimesService } from '../node-uptimes/node-uptimes.service';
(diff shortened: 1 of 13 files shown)


src/users/users.controller.ts
```diff
@@ -275,18 +282,27 @@ export class UsersController {
...
       limit,
       search,
       countryCode,
       eventTypes,
     });
+    const data = [];
+    for (const record of records) {
+      const userPoints = await this.userPointsService.findOrThrow(record.id);
+      data.push(serializedUserFromRecord(record, userPoints));
+    }
```
NullSoldier, line 301: Optionally, If you want a little potential speed boost, you could do something like this.
```suggestion
    const data = await Promise.all(records.map(async (record) => {
        const userPoints = await this.userPointsService.findOrThrow(record.id);
        return serializedUserFromRecord(record, userPoints);
    }));
```

NullSoldier (approved):

PR #816, opened 2022-05-06
Move Deposits.upsert into its own service

NullSoldier:
## Summary

We need this to avoid a circular dependency where events needs to
reference the Deposits service.

## Testing Plan

Run tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/deposits.service.ts
+++ b/src/events/deposits.service.ts
@@ -1,23 +1,17 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { Injectable } from '@nestjs/common';
-import { Deposit, DepositHead, EventType } from '@prisma/client';
+import { Injectable, NotFoundException } from '@nestjs/common';
+import { Deposit, DepositHead } from '@prisma/client';
 import { ApiConfigService } from '../api-config/api-config.service';
-import { BlockOperation } from '../blocks/enums/block-operation';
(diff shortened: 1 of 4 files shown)


src/events/deposits.service.ts
```diff
@@ -1,23 +1,17 @@
...
-import { UpsertDepositsOperationDto } from './dto/upsert-deposit.dto';
-import { EventsService } from './events.service';
 
 @Injectable()
 export class DepositsService {
   constructor(
     private readonly config: ApiConfigService,
     private readonly prisma: PrismaService,
-    private readonly events: EventsService,
     private readonly users: UsersService,
```
rohanjadvani, line 15: Can this be removed?

brekk (approved):

src/events/deposits.service.ts
NullSoldier: Removed that and config

src/events/deposits.service.ts
```diff
@@ -1,25 +1,15 @@
...
-  constructor(
-    private readonly config: ApiConfigService,
-    private readonly prisma: PrismaService,
-    private readonly events: EventsService,
-    private readonly users: UsersService,
-  ) {}
+  constructor(private readonly prisma: PrismaService) {}
 
   async find(id: number): Promise<Deposit | null> {
     return await this.prisma.deposit.findUnique({
```
rohanjadvani, line 15:
```suggestion
    return this.prisma.deposit.findUnique({
```

rohanjadvani (approved):

src/events/deposits.service.ts
NullSoldier: done

PR #825, opened 2022-05-10
no more block mined points

lwisne:
## Summary
Creates a config var for counting block mined points. Set it to true for test to minimize the overall set of changes to tests etc. Otherwise default to false. 

## Breaking Change

```
[ ] Yes
```


.env.test
```diff
@@ -1,3 +1,4 @@
+ALLOW_BLOCK_MINED_POINTS=true
```
rohanjadvani, line 1: Can this be added to [`.env.template`](https://github.com/iron-fish/ironfish-api/blob/master/.env.template) too?

src/events/events.service.spec.ts
```diff
@@ -896,6 +896,15 @@ describe('EventsService', () => {
         expect(event).toBeNull();
       });
     });
+
+    describe('respects enviornment variables', () => {
```
rohanjadvani, line 900:
```suggestion
    describe('when `ALLOW_BLOCK_MINED_POINTS` is false', () => {
```

src/events/events.service.spec.ts
```diff
@@ -896,6 +896,15 @@ describe('EventsService', () => {
...
+      it('returns null if config returns false for get ALLOW_BLOCK_MINED_POINTS', async () => {
```
rohanjadvani, line 901:
```suggestion
      it('returns null', async () => {
```

rohanjadvani (approved):

.env.test
lwisne: yup, done

PR #834, opened 2022-05-11
feat(api): Check numbers for max value before sending to Prisma

rohanjadvani:
## Summary

I love ORMs

## Testing Plan

Manual testing

## Breaking Change

```
[ ] Yes
```


brekk (approved):
Big numbers are hard

PR #838, opened 2022-05-12
Move updating user points table to graphile worker job

danield9tqh:
## Summary
The user_points aggregation is pretty slow, so it could help our API performance to move it out into a worker task.

## Testing Plan
Unit tests and testing on staging

## Breaking Change

```
[ ] Yes
```


src/events/events.jobs.controller.ts
```diff
@@ -55,4 +60,59 @@ export class EventsJobsController {
...
+
+  @MessagePattern(GraphileWorkerPattern.UPDATE_LATEST_POINTS)
+  async updateLatestPoints({
+    userId,
+    type,
+  }: {
+    userId: number;
+    type: EventType;
+  }): Promise<void> {
+    await this.prisma.$transaction(async (prisma) => {
```
rohanjadvani, line 72: We generally don't use `prisma` in our controllers to leave DB abstractions at the service level. Could we move this aggregation to the `EventsService` and call that method here?

src/events/events.service.ts
```diff
@@ -548,7 +551,13 @@ export class EventsService {
       });
     }
 
-    await this.updateLatestPoints(userId, type, occurredAt, client);
+    await this.graphileWorkerService.addJob(
+      GraphileWorkerPattern.UPDATE_LATEST_POINTS,
+      { userId: userId, type: type },
```
rohanjadvani, line 556:
```suggestion
      { userId, type },
```

rohanjadvani (changes requested):

src/events/events.jobs.controller.ts
danield9tqh: Yea I can move this there

src/events/events.service.spec.ts
```diff
@@ -513,6 +527,20 @@ describe('EventsService', () => {
         });
 
         await eventsService.create({ type, userId: user.id, points: 100 });
+        await eventsJobsController.updateLatestPoints({
```
rohanjadvani, line 530: To avoid importing the worker controller in this test

```suggestion
        await eventsService.updateLatestPoints(user.id, type)
```

src/events/events.service.spec.ts
```diff
@@ -565,6 +593,20 @@ describe('EventsService', () => {
 
         const points = 200;
         await eventsService.create({ type, userId: user.id, points });
+        await eventsJobsController.updateLatestPoints({
```
rohanjadvani, line 596:
```suggestion
        await eventsService.updateLatestPoints(user.id, type)
```

src/events/events.service.spec.ts
```diff
@@ -636,6 +678,20 @@ describe('EventsService', () => {
         userId: user.id,
         points,
       });
+      await eventsJobsController.updateLatestPoints({
```
rohanjadvani, line 681:
```suggestion
      await eventsService.updateLatestPoints(user.id, type)
```

src/events/events.service.spec.ts
```diff
@@ -761,6 +817,19 @@ describe('EventsService', () => {
         points,
         url,
       });
+      await eventsJobsController.updateLatestPoints({
```
rohanjadvani, line 820:
```suggestion
      await eventsService.updateLatestPoints(user.id, type)
```

src/events/events.service.spec.ts
```diff
@@ -865,6 +947,19 @@ describe('EventsService', () => {
       it('adds points to the user', async () => {
         const { block, user } = await setupBlockMinedWithEvent(0);
         await eventsService.upsertBlockMined(block, user);
+        await eventsJobsController.updateLatestPoints({
```
rohanjadvani, line 950:
```suggestion
        await eventsService.updateLatestPoints(user.id, type)
```

src/events/events.service.spec.ts
```diff
@@ -931,6 +1026,20 @@ describe('EventsService', () => {
         const { block, event, user } = await setupBlockMinedWithEvent();
         const currentUserPoints = await userPointsService.findOrThrow(user.id);
         await eventsService.deleteBlockMined(block);
+        await eventsJobsController.updateLatestPoints({
```
rohanjadvani, line 1029:
```suggestion
        await eventsService.updateLatestPoints(user.id, type)
```

src/events/events.service.spec.ts
```diff
@@ -956,6 +1065,20 @@ describe('EventsService', () => {
       const { event, user } = await setupBlockMinedWithEvent();
       const currentUserPoints = await userPointsService.findOrThrow(user.id);
       await eventsService.delete(event);
+      await eventsJobsController.updateLatestPoints({
```
rohanjadvani, line 1068:
```suggestion
      await eventsService.updateLatestPoints(user.id, type)
```

src/events/events.service.spec.ts
```diff
@@ -966,6 +1089,19 @@ describe('EventsService', () => {
...
```
rohanjadvani, line 1092:
```suggestion
      await eventsService.updateLatestPoints(user.id, type)
```

src/users/users.controller.spec.ts
```diff
@@ -278,6 +281,16 @@ describe('UsersController', () => {
           points: 500,
         });
 
+        await eventsJobsController.updateLatestPoints({
```
rohanjadvani, line 284:
```suggestion
        await eventsService.updateLatestPoints(user.id, type)
```

src/users/users.controller.spec.ts
```diff
@@ -278,6 +281,16 @@ describe('UsersController', () => {
...
+          userId: user.id,
+          type: EventType.PULL_REQUEST_MERGED,
+        });
+
+        await eventsJobsController.updateLatestPoints({
```
rohanjadvani, line 289:
```suggestion
        await eventsService.updateLatestPoints(user.id, type)
```

rohanjadvani (commented):
The code changes look good, but I think we should refactor the tests to use the `EventsService` instead of the `EventsJobsController`. The jobs controllers should be tested separately and aren't direct dependencies of the REST controllers (ex. `UsersController`) or the DB services (ex. `EventsService`). If we want to kick off an aggregation in a test to validate the points for a user, we should use the service method for that (`EventsService.updateLatestPoints`).

src/events/events.service.spec.ts
danield9tqh: I think it makes more sense to do `eventsJobsController.updateLatestPoints` because that is the actual function that is being called when the job is executed. Yes, ideally we would test the job separately. Then we would remove ALL of these checks for points in the current tests and just check that the job was kicked off. But that seemed like too much effort and more likelihood of making a mistake / missing some case when copying over all the expectations to the job test.

danield9tqh: Sort of like in each test
1. Check that the job was kicked off
2. Run the job synchronously and check expectations

If I could've found a way to wait for the asynchronous job to finish and then test expectations that would have been ideal. But I couldn't figure out how to do that

rohanjadvani: Yeah, I agree the ideal scenario will be removing the checks for the points and seeing the job was enqueued in the service since we're moving to eventual consistency. The job tests can test the point value changes. I'm fine leaving as is if you want to keep that refactoring out of this PR.

rohanjadvani (approved):

src/events/events.service.spec.ts
danield9tqh: 👍🏻 thanks. yea not ideal to not do all the test refactoring but I think it's probably good to get this out quickly to improve API performance

PR #847, opened 2022-05-16
Improve performance on update_latest_points graphile job

danield9tqh:
## Summary
We have a large number of update_latest_points graphile jobs queued in prod. Many of them are duplicates which we can improve by not adding duplicate jobs into the queue using a [jobKey](https://github.com/graphile/worker#replacing-updating-and-removing-jobs). Also this removes the `queueName` parameter. `queueName` is used when jobs need to be [executed synchronously in a queue](https://github.com/graphile/worker#parsedcronitems). For these jobs we do not require that so maybe removing this will help speed them up as well.

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -24,6 +24,7 @@ import { EventWithMetadata } from './interfaces/event-with-metadata';
 import { ListEventsOptions } from './interfaces/list-events-options';
 import { SerializedEventMetrics } from './interfaces/serialized-event-metrics';
 import { Block, Event, EventType, Prisma, User } from '.prisma/client';
+import { Job } from 'graphile-worker';
 
 // 2021 December 1 8 PM UTC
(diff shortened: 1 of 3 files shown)


src/events/events.service.ts
```diff
@@ -551,20 +552,28 @@ export class EventsService {
...
+  private addUpdateLatestPointsJob(
+    userId: number,
+    type: EventType,
+  ): Promise<Job> {
+    const jobKey = 'ulp:' + userId.toString() + ':' + type.toString();
+    return this.graphileWorkerService.addJob(
+      GraphileWorkerPattern.UPDATE_LATEST_POINTS,
+      { userId, type },
+      {
+        jobKey,
```
rohanjadvani, line 572:
```suggestion
        jobKey: `ulp:${userId}:${type}`,
```

rohanjadvani (approved):
I think you might need to run `$ yarn lint:fix`, but otherwise looks good.

PR #851, opened 2022-05-16
feat(api): Use `user_points` when fetching event type ranks

rohanjadvani:
## Summary

Use the materialized table for ranking instead of performing aggregations on `events`.

## Testing Plan

Manual testing + staging testing

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/events.service.spec.ts
+++ b/src/events/events.service.spec.ts
@@ -1146,55 +1146,49 @@ describe('EventsService', () => {
 
       const now = new Date();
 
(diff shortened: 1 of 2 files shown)


lwisne:
I am not familiar with the method here - this is an aggregation query with a materialized view, but I don't see the latter?

lwisne (approved):
lgtm - as discussed in Slack, we could cache the aggregation query in a user_points table if we are still seeing performance issues after this.

PR #859, opened 2022-05-18
take this upsert out of an unnecessary transaction

lwisne:
## Summary
We had a transaction around a mass upsert that would be blocking some important tables and would also cause the syncer to fail if one of the transactions is invalid. Both of those seem bad, so taking it out.

(This change is mostly whitespace, be sure to view with the flag on)

Fixes IRO-1986

## Testing Plan
automated tests pass

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/blocks-transactions-loader/blocks-transactions-loader.ts
+++ b/src/blocks-transactions-loader/blocks-transactions-loader.ts
@@ -31,76 +31,66 @@ export class BlocksTransactionsLoader {
     const deleteBlockMinedPayloads: DeleteBlockMinedEventOptions[] = [];
     const upsertBlockMinedPayloads: UpsertBlockMinedEventOptions[] = [];
 
-    const response = await this.prisma.$transaction(
-      async (prisma) => {
-        const records: (Block & { transactions: Transaction[] })[] = [];
-        const previousHashes = new Map<string, BlockDto>();
-        for (const block of blocks) {
-          previousHashes.set(block.hash, block);
-        }
+    const response: (Block & { transactions: Transaction[] })[] = [];
+    const previousHashes = new Map<string, BlockDto>();
+    for (const block of blocks) {
+      previousHashes.set(block.hash, block);
+    }
 
-        for (const block of blocks) {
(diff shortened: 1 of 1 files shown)


src/blocks-transactions-loader/blocks-transactions-loader.ts
```diff
@@ -31,76 +31,66 @@ export class BlocksTransactionsLoader {
     const deleteBlockMinedPayloads: DeleteBlockMinedEventOptions[] = [];
     const upsertBlockMinedPayloads: UpsertBlockMinedEventOptions[] = [];
 
-    const response = await this.prisma.$transaction(
-      async (prisma) => {
```
rohanjadvani, line 35: Do you think we should have a transaction per block in this upsert loop to make operations for each block and associated transactions atomic?

lwisne: as discussed, added a comment to address

rohanjadvani (approved):

rohanjadvani:
Looking into these changes more and the p95 for POST `/blocks`, I don't think the Prisma transaction should be removed altogether. The tables that are affected by this flow are `blocks`, `blocks_transactions`, and `transactions` (+ `blocks_daily` via a worker job). Users would not experience any slowness from this, only the syncer and as a result latency in the block explorer. 

If the syncer has a failure, we have an alert for this via Datadog. If there is an invalid record within transactions for a block and the Prisma transaction is no longer present, the API would persist incomplete data and the explorer would render the invalid data. 

If any immediate refactor should be done here, I think it should be moving the Prisma transaction to be per block (which is still a good change to move away from unnecessary loops within this transaction). The more I think about this, the less I want to compromise atomicity and consistency since this request is not a high priority bottleneck. If there is an issue within the bulk upsert, the syncer will pick up the most recent head and retry going forward.

lwisne:
OK. Respectfully disagree here - I don't feel comfortable putting a loop with 600 upserts into a `$transaction` block. 

I think doing it by block would be a half-measure, so will just close for now.

PR #863, opened 2022-05-24
feat(api): Speed up `POST /deposits`

rohanjadvani:
## Summary

* Update bulk upsert to update the deposit head and enqueue Graphile jobs
* Update `POST /deposits` to return `ACCEPTED` instead of `CREATED` (response is unused)
* Moved deposit heads into module
* Remove now unused `SerializedDeposit` type

## Testing Plan

Refactored existing unit tests and added new unit tests.

## Breaking Change

```
[ ] Yes
```


src/events/deposits.upsert.service.ts
```diff
@@ -34,14 +35,21 @@ export class DepositsUpsertService {
...
-        }
+        await this.graphileWorkerService.addJob<UpsertDepositsOperationDto>(
+          GraphileWorkerPattern.UPSERT_DEPOSIT,
+          operation,
+          {
+            jobKey: `upsert_deposit:${operation.block.hash}:${operation.type}`,
+          },
+        );
+
+        const headHash =
```
mat-if, line 46: Decoupling updating the headHash from the actual deposit upsert makes me a little bit nervous, as it opens an avenue where if a job fails for any particular reason, it won't get re-processed. I may be making a mountain out of a molehill here, though

mat-if, line 46: Decoupling updating the headHash from the actual deposit upsert makes me a little bit nervous, as it opens an avenue where if a job fails for any particular reason, it won't get re-processed. I may be making a mountain out of a molehill here, though

rohanjadvani: Resolving one comment since there is a duplicate.

rohanjadvani: I think it's low risk, but that's a fair point. Moved the head hash update within the upsert.

src/events/deposits.jobs.controller.ts
```diff
@@ -0,0 +1,22 @@
...
+
+@Controller()
+export class DepositsJobsController {
+  constructor(private readonly depositsUpsertService: DepositsUpsertService) {}
+
+  @MessagePattern(GraphileWorkerPattern.UPSERT_DEPOSIT)
+  async upsert(
+    options: UpsertDepositsOperationDto,
+  ): Promise<GraphileWorkerHandlerResponse> {
+    await this.depositsUpsertService.upsert(options);
```
lwisne, line 19: why await?

src/events/deposits.controller.ts
```diff
@@ -62,15 +63,9 @@ export class DepositsController {
...
-    return deposits.map((d) => ({
-      id: d.id,
-      transaction_hash: d.transaction_hash,
-      block_sequence: d.block_sequence,
-      block_hash: d.block_hash,
-      object: 'deposit',
-    }));
+    @Res() res: Response,
+  ): Promise<void> {
+    await this.depositsUpsert.bulkUpsert(data.operations);
```
lwisne, line 68: why await?

lwisne (commented):
left a few comments about use of `await` - I had noticed we almost always use it, seems that there are some places where we could fire off the async call and move on?

src/events/deposits.controller.ts
rohanjadvani: We need an `await` here to propagate exceptions and errors properly to the NestJS [execution context](https://github.com/nestjs/nest/blob/master/packages/core/router/router-execution-context.ts#L139-L149). If an error happened without this `await`, the cluster would [exit](https://github.com/iron-fish/ironfish-api/blob/master/src/main.ts#L116-L122) and re-fork. With the `await`, NestJS properly sees exceptions and throws a 500.

Additionally, having these `await`s helps with timing and monitoring for the statistics we send over to Datadog.

src/events/deposits.jobs.controller.ts
rohanjadvani: This is in a worker task, so removing the `await` wouldn't matter too much. Additionally it helps us monitor how each task is performing and how long execution takes.

mat-if (approved):

PR #869, opened 2022-05-25
feat(api): Add read client for ranking queries

rohanjadvani:
## Summary

There is a read replica now set up on production. This code change uses a new client to perform aggregations and rankings to reduce CPU load on the main machine.

## Testing Plan

Manually ran against read replica URL.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/api-config/api-config.service.ts
+++ b/src/api-config/api-config.service.ts
@@ -31,4 +31,18 @@ export class ApiConfigService {
   isProduction(): boolean {
     return this.get<string>('NODE_ENV') === 'production';
   }
+
+  get dbPoolUrl(): string {
+    return `${this.get<string>('DATABASE_CONNECTION_POOL_URL')}?pgbouncer=true`;
+  }
+
+  get readDbPoolUrl(): string {
+    const readDbPoolUrl = this.config.get<string>(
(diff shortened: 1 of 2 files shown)


src/api-config/api-config.service.ts
```diff
@@ -31,4 +31,18 @@ export class ApiConfigService {
   isProduction(): boolean {
     return this.get<string>('NODE_ENV') === 'production';
   }
+
+  get dbPoolUrl(): string {
+    return `${this.get<string>('DATABASE_CONNECTION_POOL_URL')}?pgbouncer=true`;
+  }
+
+  get readDbPoolUrl(): string {
+    const readDbPoolUrl = this.config.get<string>(
```
mat-if, line 40: I believe config.get throws if the key isn't found (although it probably shouldn't), should this be like a try/catch or something?

mat-if (approved):

src/api-config/api-config.service.ts
rohanjadvani: This is using the build-in NestJS Config service which does not appear to throw, only [return `undefined`](https://github.com/nestjs/config/blob/master/lib/config.service.ts#L98-L123).

PR #882, opened 2022-06-01
POST /blocks now upserts incrementally

mat-if:
## Summary

The actual change is just to move the DB transaction down into the block for-loop, so that it won't rollback the entire operation if it times out

The intended effect here is that even if the request times out, it will still progress on the ones that were successfully upserted, such that it will eventually process them all.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/blocks-transactions-loader/blocks-transactions-loader.ts
+++ b/src/blocks-transactions-loader/blocks-transactions-loader.ts
@@ -2,7 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { Injectable } from '@nestjs/common';
-import { Block, Transaction } from '@prisma/client';
 import { BlocksService } from '../blocks/blocks.service';
 import { BlockDto, UpsertBlocksDto } from '../blocks/dto/upsert-blocks.dto';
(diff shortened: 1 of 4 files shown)


src/blocks/blocks.controller.ts
```diff
@@ -62,16 +62,9 @@ export class BlocksController {
...
-    );
-    return {
-      object: 'list',
-      data: blocks.map((block) =>
-        serializedBlockFromRecordWithTransactions(block),
-      ),
-    };
+  ): Promise<void> {
+    await this.blocksTransactionsLoader.bulkUpsert(upsertBlocksDto);
+    return;
```
rohanjadvani, line 67: What do you think about leaving the response like before? I have found it useful before for local development and don't think we'd lose anything.

src/blocks-transactions-loader/blocks-transactions-loader.ts
```diff
@@ -129,6 +123,6 @@ export class BlocksTransactionsLoader {
       },
     );
 
-    return response;
+    return;
```
brekk, line 126: This isn't needed?

brekk (commented):
Minor suggestions, probably not blockers

rohanjadvani (approved):

PR #920, opened 2022-06-21
Expose health/admin with queued job count

NullSoldier:
## Summary

We can use this in data dog to set up alerts for when our queued job
count gets too high.

## Testing Plan

Wrote a test, but also run the API and hit the API route. `http://localhost:8003/health/admin`

## Breaking Change

```
[ ] Yes
```


src/graphile-worker/graphile-worker.service.ts
```diff
@@ -30,6 +30,20 @@ export class GraphileWorkerService {
...
   }
 
+  async queuedJobCount(): Promise<number> {
+    if (!this.workerUtils) {
+      await this.initWorkerUtils();
+    }
+
+    return this.workerUtils.withPgClient(async (pgClient) => {
+      const result = await pgClient.query<{ count: string }>(
+        'select count(*) from graphile_worker.jobs where locked_at IS NULL;',
```
rohanjadvani, line 40:
```suggestion
        'SELECT COUNT(*) FROM graphile_worker.jobs WHERE locked_at IS NULL;',
```

src/health/health.controller.ts
```diff
@@ -1,15 +1,30 @@
...
   }
+
+  @ApiExcludeEndpoint()
+  @UseGuards(ApiKeyGuard)
+  @Get('admin')
+  async admin(): Promise<{
+    jobQueueCount: number;
+  }> {
+    return {
+      jobQueueCount: await this.graphileService.queuedJobCount(),
```
rohanjadvani, line 27: nit: We generally use snake-case for fields returned by our API.

rohanjadvani: What do you think about `queued_jobs` (also more consistent with `queuedJobCount`)?
```suggestion
      queued_jobs: await this.graphileService.queuedJobCount(),
```

src/health/health.controller.ts
```diff
@@ -1,15 +1,30 @@
...
   @Get()
   health(): string {
     return 'OK';
   }
+
+  @ApiExcludeEndpoint()
+  @UseGuards(ApiKeyGuard)
+  @Get('admin')
+  async admin(): Promise<{
+    jobQueueCount: number;
```
rohanjadvani, line 24: I think we should update this return type now.

rohanjadvani:
```suggestion
    queued_jobs: number;
```

PR #925, opened 2022-06-21
Fix graphile WorkerUtils race condition

NullSoldier:
## Summary

There is a race condition here if initWorkerUtils() is called multiple
times in parallel.

## Testing Plan

Run tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/graphile-worker/graphile-worker.service.ts
+++ b/src/graphile-worker/graphile-worker.service.ts
@@ -12,6 +12,7 @@ import { GraphileWorkerJobOptions } from './interfaces/graphile-worker-job-optio
 @Injectable()
 export class GraphileWorkerService {
   private workerUtils!: WorkerUtils;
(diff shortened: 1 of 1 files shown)


deekerno (approved):

src/graphile-worker/graphile-worker.service.ts
```diff
@@ -45,9 +43,19 @@ export class GraphileWorkerService {
   }
 
   private async initWorkerUtils(): Promise<void> {
-    this.workerUtils = await makeWorkerUtils({
+    if (this.workerUtils) {
+      return;
+    }
+
+    if (this.workerUtilsPromise) {
+      await this.workerUtilsPromise;
```
rohanjadvani, line 51: This will fall through and re-assign. We should return in this conditional.

PR #926, opened 2022-06-22
refactor: move INFLUXDB_ORG to config

wd021:
## Summary
add INFLUXDB_ORG to config instead of hardcoding in `influxdb.service.ts`

## Testing Plan
- add INFLUXDB_ORG to env var and restart api

## Breaking Change

```
[ ] Yes
```


wd021:
@rohanjadvani updated all

rohanjadvani (approved):

PR #927, opened 2022-06-22
refactor: change telemetry timestamp type

wd021:
## Summary
Telemetry api is currently using `Date` obj for the timestamp. We should update this to `number` and have millisecond-level precision instead of seconds.

## Testing Plan
- Run Tests
- Restart InfluxDB
- Update `ironfish` version
 
## Breaking Change

```
[x] Yes
- requires updating ironfish node version to support changes in telemetry api
```


rohanjadvani:
Hey @wd021 - I'm going to close this PR. I need to add a contributing doc (similar to the [one](https://github.com/iron-fish/ironfish/blob/master/CONTRIBUTING.md) in our main repo), but for a core change to the telemetry system, it'd be great to have an issue and proposed changes opened up first. This level of granularity won't strictly improve the existing logic, and we'd need some logic changed both up and downstream.

wd021:
@rohanjadvani gotcha. yea wasn't sure if this kind of change qualifies as needing a FIP. I'll close https://github.com/iron-fish/ironfish/pull/1664 in the main repo as well since it's linked to this.

PR #940, opened 2022-06-29
Create endpoint to upload peer list

danield9tqh:
## Summary
Create an endpoint in telemetry for nodes to upload peer lists. This will be used to create a network topology graph

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/telemetry/dto/write-peer-list.dto.ts
+++ b/src/telemetry/dto/write-peer-list.dto.ts
@@ -1,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { Type } from 'class-transformer';
+import {
+  ArrayMaxSize,
+  IsArray,
+  IsDate,
+  IsNotEmpty,
+  IsString,
+  MaxLength,
+} from 'class-validator';
+
+export class WritePeerListDto {
+  @IsArray()
+  @ArrayMaxSize(10000)
+  @MaxLength(100, {
+    each: true,
+  })
+  readonly peers!: string[];
+
+  @IsString()
+  @IsNotEmpty()
+  @MaxLength(100)
+  readonly nodeId!: string;
+
+  @IsDate()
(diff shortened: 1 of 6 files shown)


src/telemetry/telemetry.controller.ts
```diff
@@ -90,6 +93,36 @@ export class TelemetryController {
...
+      return this.prisma.peerConnection.upsert({
+        create: upload,
+        update: upload,
+        where: {
+          uq_peer_connection_on_source_dest_time: upload,
+        },
+      });
+    });
+
+    await Promise.all(promises);
```
rohanjadvani, line 123: We shouldn't use a `Promise.all` on these payloads since that can lead to starvation for the DB.

src/telemetry/dto/write-peer-list.dto.ts
```diff
@@ -0,0 +1,31 @@
...
+  @ArrayMaxSize(10000)
+  @MaxLength(100, {
+    each: true,
+  })
+  readonly peers!: string[];
+
+  @IsString()
+  @IsNotEmpty()
+  @MaxLength(100)
+  readonly nodeId!: string;
```
rohanjadvani, line 26: We use snake case for our API payloads

```suggestion
  readonly node_id!: string;
```

rohanjadvani (changes requested):
I think it'd be good to chunk out this PR into several smaller changes:
* Migration for new table
* Service to create/upsert records
* Controller to accept requests

What do you think?

src/telemetry/telemetry.controller.ts
```diff
@@ -90,6 +93,36 @@ export class TelemetryController {
...
+        sourceId: nodeId,
+        destinationId: peer,
+        timestamp: timestamp,
+      };
+
+      return this.prisma.peerConnection.upsert({
+        create: upload,
+        update: upload,
+        where: {
+          uq_peer_connection_on_source_dest_time: upload,
```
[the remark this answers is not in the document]
danield9tqh: No, not intentionally. I did want to keep them unique on these 3 fields. If that's the case is there a better option than `upsert` here?

danield9tqh:
closing this because we decided to use telemetry / influxdb to upload the peer list instead

PR #947, opened 2022-07-06
exposes mismatched deposits count in admin health check

hughy:
## Summary

- queries deposits and blocks tables to count deposits where either the block
  hash does not exist in blocks or there is a mismatch in 'main'
- adds DepositUpsertService to HealthController to expose mismatched deposits
  count

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/deposits.upsert.service.ts
+++ b/src/events/deposits.upsert.service.ts
@@ -139,9 +139,9 @@ export class DepositsUpsertService {
       SELECT
         COUNT(*)
       FROM
-        blocks
-      RIGHT OUTER JOIN
         deposits
+      LEFT JOIN
+        blocks
       ON blocks.hash = deposits.block_hash
       WHERE blocks.hash IS NULL OR blocks.main <> deposits.main
       `,
--- a/src/events/deposits.upsert.service.ts
+++ b/src/events/deposits.upsert.service.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { Injectable } from '@nestjs/common';
 import { Deposit, EventType } from '@prisma/client';
+import is from '@sindresorhus/is';
 import { ApiConfigService } from '../api-config/api-config.service';
 import { BlockOperation } from '../blocks/enums/block-operation';
 import { SEND_TRANSACTION_LIMIT_ORE } from '../common/constants';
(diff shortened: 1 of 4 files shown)


src/events/deposits.upsert.service.ts
```diff
@@ -131,4 +132,23 @@ export class DepositsUpsertService {
...
+      SELECT
+        COUNT(*) AS count
+      FROM
+        blocks
+      RIGHT OUTER JOIN
+        deposits
+      ON blocks.hash = deposits.block_hash
+      WHERE blocks.hash IS NULL OR blocks.main <> deposits.main
+      `,
+    );
```
NullSoldier, line 148: We should figure out how to run this against the read only slave.

NullSoldier:
```
Finalize Aggregate  (cost=312701.30..312701.30 rows=1 width=8)
   ->  Gather  (cost=312701.20..312701.30 rows=1 width=8)
         Workers Planned: 1
         ->  Partial Aggregate  (cost=311701.20..311701.20 rows=1 width=8)
               ->  Parallel Hash Left Join  (cost=3870.07..311022.97 rows=1356454 width=0)
                     Hash Cond: ((deposits.block_hash)::text = (blocks.hash)::text)
                     Filter: ((blocks.hash IS NULL) OR (blocks.main <> deposits.main))
                     ->  Parallel Seq Scan on deposits  (cost=0.00..290953.20 rows=5767399 width=66)
                     ->  Parallel Hash  (cost=3661.11..3661.11 rows=59704 width=66)
                           ->  Parallel Seq Scan on blocks  (cost=0.00..3661.11 rows=59704 width=66)
```

hughy: It looks like `queryRawUnsafe` is already set to query from the read instance on production https://github.com/iron-fish/ironfish-api/blob/master/src/prisma/prisma.service.ts#L33

src/events/deposits.upsert.service.ts
```diff
@@ -131,4 +132,23 @@ export class DepositsUpsertService {
 
     return deposits;
   }
+
+  async mismatchedDepositCount(): Promise<number> {
+    const result = await this.prisma.$queryRawUnsafe<{ count: number }[]>(
+      `
+      SELECT
+        COUNT(*) AS count
```
rohanjadvani, line 140: nit: I think you can omit `AS count`, as Postgres defaults `COUNT` to this column

src/health/health.controller.ts
```diff
@@ -4,12 +4,16 @@
...
 import { ApiKeyGuard } from '../auth/guards/api-key.guard';
+import { DepositsUpsertService } from '../events/deposits.upsert.service';
 import { GraphileWorkerService } from '../graphile-worker/graphile-worker.service';
 
 @ApiTags('Health')
 @Controller('health')
 export class HealthController {
-  constructor(private readonly graphileService: GraphileWorkerService) {}
+  constructor(
+    private readonly graphileService: GraphileWorkerService,
```
rohanjadvani, line 14:
```suggestion
    private readonly graphileWorkerService: GraphileWorkerService,
```

rohanjadvani (approved):
Looks good! I have a couple of nits, but feel free to merge without addressing. They are opinionated.

PR #948, opened 2022-07-06
outlines queries to fix data from out of order deposits

hughy:
## Summary

backfill_mismatched_deposits.sql includes four queries to do the following:

1. soft-delete events from deposits that don't correspond to a block on the main
chain. ('overcounted' deposits)
2. upsert events from deposits on the main chain but currently marked otherwise
('undercounted' deposits)
3. set deposits.main to match blocks.main for any deposits where it doesn't
4. set deposits.main to false for any deposit marked main that doesn't have a
block in blocks

NOTE: this PR isn't necessarily intended to be merged into the repo, but useful for sharing and reviewing the queries before we run them.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/backfill_mismatched_deposits.sql
+++ b/backfill_mismatched_deposits.sql
@@ -1,0 +1,64 @@
+-- soft-delete events for overcounted deposits
+UPDATE events SET deleted_at = current_timestamp
+WHERE
+  deposit_id IN (
+  SELECT 
+    deposits.id
+  FROM
+    deposits
+  LEFT JOIN 
+    blocks
+  ON
+    deposits.block_hash = blocks.hash
+  WHERE
+    deposits.main = true AND (blocks.main IS NULL OR blocks.main = false)
+);
+
+-- upsert events for undercounted mismatched deposits
+INSERT INTO events (type, occurred_at, points, user_id, deposit_id)
+SELECT
+  'SEND_TRANSACTION' AS type,
+  missing_events.block_timestamp AS occurred_at,
+  1 AS points,
+  users.id AS user_id,
+  missing_events.deposit_id AS deposit_id
+FROM
+  users
+JOIN
+  (
+    SELECT
+      deposits.id AS deposit_id,
+      deposits.graffiti AS deposit_graffiti
+      blocks.timestamp AS block_timestamp
+    FROM
+      deposits
+    JOIN
+      blocks
+    ON
+      deposits.block_hash = blocks.hash
+    WHERE
+      deposits.main = false AND blocks.main = true AND deposits.amount >= 10000000
+  ) missing_events
+ON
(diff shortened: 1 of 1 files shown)


backfill_mismatched_deposits.sql
```diff
@@ -0,0 +1,64 @@
...
+  SELECT 
+    deposits.id
+  FROM
+    deposits
+  LEFT JOIN 
+    blocks
+  ON
+    deposits.block_hash = blocks.hash
+  WHERE
+    deposits.main = true AND (blocks.main IS NULL OR blocks.main = false)
```
danield9tqh, line 14: Do we know a case where blocks.main would be null?

backfill_mismatched_deposits.sql
```diff
@@ -0,0 +1,64 @@
+-- soft-delete events for overcounted deposits
+UPDATE events SET deleted_at = current_timestamp
```
danield9tqh, line 2: Do we also need to set points to 0 here? It looks like we do that [when 'deleting' deposits](https://github.com/iron-fish/ironfish-api/blob/61b1cde0b82e412ab691ddc90c5e9e2ff2c27cac/src/events/events.service.ts#L619)

backfill_mismatched_deposits.sql
```diff
@@ -0,0 +1,64 @@
...
+  LEFT JOIN 
+    blocks
+  ON
+    deposits.block_hash = blocks.hash
+  WHERE
+    deposits.main = true AND (blocks.main IS NULL OR blocks.main = false)
+);
+
+-- upsert events for undercounted mismatched deposits
+INSERT INTO events (type, occurred_at, points, user_id, deposit_id)
```
danield9tqh, line 18: Just want to make sure that this updates instead of creating new events. Because we have the [event.id as auto-increment](https://github.com/iron-fish/ironfish-api/blob/61b1cde0b82e412ab691ddc90c5e9e2ff2c27cac/prisma/schema.prisma#L12). Was wondering if `UPDATE` maybe isn't better for this query since I can't think of scenarios where the event would not already be in the table

backfill_mismatched_deposits.sql
hughy: There are some cases (~1.5k) where there's no record in `blocks` that matches `deposits.block_hash`, so this is meant to capture those.

backfill_mismatched_deposits.sql
hughy: Ah good call, there needs to be another clause in this query with `ON CONFLICT` to address updates. There's a uniqueness constraint on `deposit_id`, so this will only insert events for deposits where there isn't one already.

backfill_mismatched_deposits.sql
hughy: Will update!

PR #949, opened 2022-07-07
adds an api endpoint and graphile job to fix mismatched deposits

hughy:
## Summary

fixes mismatched deposits in a series of four queries:
1. soft-deletes events for deposits on forks or without a block in 'blocks'
2. upserts events for deposits on main chain but marked otherwise
3. sets deposits.main equal to blocks.main wherever mismatched
4. sets deposits.main to false for any deposits without a block in 'blocks'

a POST to '/deposits/fix_mismatches' enqueues a worker job to execute these queries.

## Testing Plan

added tests and tested locally

## Breaking Change

```
[ ] Yes
```


src/events/deposits.upsert.service.ts
```diff
@@ -151,4 +151,96 @@ export class DepositsUpsertService {
     }
     return result[0].count;
   }
+
+  async enqueueFixMismatchedDeposits(): Promise<void> {
+    await this.graphileWorkerService.addJob(
+      GraphileWorkerPattern.FIX_MISMATCHED_DEPOSITS,
+    );
+  }
```
NullSoldier, line 159: nit: I think we can move this directly into the controller, like we do for refresh_points. I don't think we need this method.

src/events/deposits.upsert.service.ts
```diff
@@ -151,4 +165,55 @@ export class DepositsUpsertService {
     }
     return result[0].count;
   }
+
+  async mismatchedDeposits(beforeSequence: number = 0): Promise<
+    (Deposit & { block_main: boolean | null; block_timestamp: Date | null })[]
+  > {
+    const blocksHead = await this.blocksService.head();
+    return await this.prisma.$queryRawUnsafe<
```
rohanjadvani, line 173:
```suggestion
    return this.prisma.$queryRawUnsafe<
```

src/events/deposits.upsert.service.ts
```diff
@@ -151,4 +165,55 @@ export class DepositsUpsertService {
...
+        const updatedDeposit = await this.prisma.deposit.update({
+          data: {
+            main: deposit.block_main ?? false,
+          },
+          where: {
+            id: deposit.id,
+          },
+        });
+
+        await this.processDeposit(
```
rohanjadvani, line 211: I think this is fine for now, but if we see any starvation or long running worker jobs maybe we can chunk this out down the line.

rohanjadvani (approved):
LGTM after lint is fixed

src/events/deposits.controller.ts
```diff
@@ -68,4 +71,13 @@ export class DepositsController {
     await this.depositsUpsert.bulkUpsert(data.operations);
     res.sendStatus(HttpStatus.ACCEPTED);
   }
+
+  @ApiExcludeEndpoint()
+  @Post('fix_mismatches')
```
rohanjadvani, line 76: What do you think about `refresh`, for consistency?

hughy: I like it, I'll change the endpoint and rename the associated methods

src/events/deposits.upsert.service.ts
hughy: I think it may be worth breaking this loop into separate jobs now -- there are over 80k mismatched deposits right now, so the first time we run this will almost surely be a long running job.

rohanjadvani (commented):
This code is great, but I think we should add some unit tests before merging for the new service methods.

PR #951, opened 2022-07-07
Preserve transaction index and sort by them

NullSoldier:
## Summary

Transactions should be returned in index order to be correct. This now
stores the index, and results them in index order.

## Testing Plan
Added tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/blocks-transactions/blocks-transactions.service.spec.ts
+++ b/src/blocks-transactions/blocks-transactions.service.spec.ts
@@ -133,7 +133,7 @@ describe('BlocksTransactionsService', () => {
               spends,
             },
           });
-          await blocksTransactionsService.upsert(prisma, block, transaction, 0);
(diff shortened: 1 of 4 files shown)


src/blocks-transactions/blocks-transactions.service.spec.ts
```diff
@@ -131,7 +133,7 @@ describe('BlocksTransactionsService', () => {
               spends,
             },
           });
-          await blocksTransactionsService.upsert(prisma, block, transaction);
+          await blocksTransactionsService.upsert(prisma, block, transaction, 0);
```
rohanjadvani, line 136:
```suggestion
          await blocksTransactionsService.upsert(prisma, block, transaction, i);
```

rohanjadvani (approved):

PR #956, opened 2022-07-11
adds dd-trace instrumentation to deposit upserts

hughy:
## Summary

- avoids repeated calls to standardizeHash for block and transaction hashes
- exports tracer singleton from tracer module
- adds tracing to each query in DepositUpsertService.upsert

## Breaking Change

```
[ ] Yes
```


src/events/deposits.upsert.service.ts
```diff
@@ -114,33 +123,51 @@ export class DepositsUpsertService {
     blockTimestamp: Date | null,
   ): Promise<void> {
     if (!deposit.main) {
-      const event = await prisma.event.findUnique({
-        where: {
-          deposit_id: deposit.id,
+      const event = await tracer.trace(
+        'DepositUpserService.processDeposit.find_event',
```
rohanjadvani, line 127:
```suggestion
        'DepositUpsertService.processDeposit.find_event',
```

src/events/deposits.upsert.service.ts
```diff
@@ -114,33 +123,51 @@ export class DepositsUpsertService {
...
+            },
+          });
         },
-      });
+      );
+
       if (event) {
-        await this.eventsService.deleteWithClient(event, prisma);
+        await tracer.trace(
+          'DepositUpsertService.processDeposit.delete_event',
```
rohanjadvani, line 139: nit: It's reads a little weird to have mixed camelCase and snake_case

src/events/deposits.upsert.service.ts
```diff
@@ -73,25 +77,30 @@ export class DepositsUpsertService {
...
-            update: depositParams,
-            where: {
-              uq_deposits_on_transaction_hash_and_graffiti: {
-                transaction_hash: depositParams.transaction_hash,
-                graffiti: depositParams.graffiti,
-              },
+          const deposit: Deposit = await tracer.trace(
+            'DepositUpsertService.upsert.upsert_deposit',
+            async (): Promise<Deposit> => {
+              return await prisma.deposit.upsert({
```
rohanjadvani, line 92:
```suggestion
              return prisma.deposit.upsert({
```

src/events/deposits.upsert.service.ts
```diff
@@ -114,33 +123,51 @@ export class DepositsUpsertService {
     blockTimestamp: Date | null,
   ): Promise<void> {
     if (!deposit.main) {
-      const event = await prisma.event.findUnique({
-        where: {
-          deposit_id: deposit.id,
+      const event = await tracer.trace(
+        'DepositUpserService.processDeposit.find_event',
+        async () => {
```
rohanjadvani, line 128: Above we use a return type. Can we add one here?

src/events/deposits.upsert.service.ts
```diff
@@ -114,33 +123,51 @@ export class DepositsUpsertService {
...
+          });
         },
-      });
+      );
+
       if (event) {
-        await this.eventsService.deleteWithClient(event, prisma);
+        await tracer.trace(
+          'DepositUpsertService.processDeposit.delete_event',
+          async () => {
```
rohanjadvani, line 140: Same comment re: return type

src/events/deposits.upsert.service.ts
```diff
@@ -114,33 +123,51 @@ export class DepositsUpsertService {
...
       }
     }
 
     if (deposit.main && deposit.amount >= SEND_TRANSACTION_LIMIT_ORE) {
-      const user = await this.usersService.findByGraffiti(
-        deposit.graffiti,
-        prisma,
+      const user = await tracer.trace(
+        'DepositUpsertService.processDeposit.find_user',
+        async () => {
```
rohanjadvani, line 150: Same comment re: return type

rohanjadvani (commented):
What's the testing plan? Are we going to deploy to staging and observe some test traces there before pushing to production?

hughy:
> What's the testing plan? Are we going to deploy to staging and observe some test traces there before pushing to production?

Yeah I think we'll need to observe in staging first. I'm not sure if there's a way to get traces on a local app into DataDog

src/events/deposits.upsert.service.ts
hughy: it does look odd! since these names are only for our benefit I'll stick to camelCase

rohanjadvani:
> > What's the testing plan? Are we going to deploy to staging and observe some test traces there before pushing to production?
> 
> Yeah I think we'll need to observe in staging first. I'm not sure if there's a way to get traces on a local app into DataDog



> > What's the testing plan? Are we going to deploy to staging and observe some test traces there before pushing to production?
> 
> Yeah I think we'll need to observe in staging first. I'm not sure if there's a way to get traces on a local app into DataDog

Sounds good; I'll update the description

rohanjadvani (approved):
I think this is fine to deploy for now. If we understand bottlenecks from this flow, then I think we can remove this, or ideally inject some tracing calls before any service methods are called.

PR #963, opened 2022-07-15
throttles UPDATE_LATEST_POINTS for each job key

hughy:
## Summary

a large percentage of `UPDATE_LATEST_POINTS` jobs share the same jobKey. the
jobKey prevents duplicate jobs from being queued, but jobs for some users and
event types are queued so frequently that there is almost always one in the
queue.

- sets runAt to defer job execution by 10 minutes
- adds jobKeyMode argument to addJob
- uses jobKeyMode=preserve_run_at to throttle jobs with the same jobKey

## Testing Plan

- updated unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -521,13 +521,20 @@ export class EventsService {
     type: EventType,
   ): Promise<Job> {
     const updateLatestPointsQueues = 4;
+    const maxJobFrequencyMinutes = 10;
+
     const queueNumber = Math.floor(Math.random() * updateLatestPointsQueues);
+    const runAt = new Date();
+    runAt.setMinutes(runAt.getMinutes() + maxJobFrequencyMinutes);
+
     return this.graphileWorkerService.addJob(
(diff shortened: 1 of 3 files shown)


src/events/events.service.ts
```diff
@@ -521,13 +521,20 @@ export class EventsService {
...
+    runAt.setMinutes(runAt.getMinutes() + maxJobFrequencyMinutes);
+
     return this.graphileWorkerService.addJob(
       GraphileWorkerPattern.UPDATE_LATEST_POINTS,
       { userId, type },
       {
         jobKey: `ulp:${userId}:${type}`,
+        jobKeyMode: `preserve_run_at`,
         queueName: `update_latest_points_${queueNumber}`,
+        runAt: runAt,
```
rohanjadvani, line 537:
```suggestion
        runAt,
```

rohanjadvani (approved):

PR #976, opened 2022-07-22
Move influx write after node uptime processing

NullSoldier:
## Summary

It's higher priority and these requests have been failing because of
influx write errors.

## Testing Plan

Run normal tests

## Breaking Change

```
[ ] Yes
```


mat-if (approved):

rohanjadvani (approved):
Looks good after the lint fix

PR #977, opened 2022-07-23
Fix performance of Events list

Skakruk:
## Summary

I am fixing an issue with API performance on the Users page as described in https://github.com/iron-fish/website-testnet/issues/271 for users with a large number of events.

Cursor paginating (in Prisma implementation) is becoming very heavy for many entries since it fetches all available rows from DB and limits the result on the client side.
Proposed changes query DB almost in the same way cursor pagination is intended to work, including `limit` in the SQL query.

See performance for `/events` route:

Before changes - ~6sec:
Screen Shot 2022-07-23 at 19 59 34

After - ~440ms:
Screen Shot 2022-07-23 at 20 00 14

UPD:
`/metrics` route also was performing poorly on a big amount of events:
Before:
Screen Shot 2022-07-26 at 12 55 30

After:
Screen Shot 2022-07-26 at 13 06 55

## Testing Plan

Tests are passing without changes

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -64,13 +64,7 @@ export class EventsService {
       occurred_at: Prisma.SortOrder.desc,
     };
     const skip = cursorId ? 1 : 0;
-    const where: {
-      user_id?: number;
-      deleted_at: null;
-      id?: {
-        [key: string]: number;
-      };
-    } = {
+    const where: Prisma.EventWhereInput = {
       user_id: options.userId,
       deleted_at: null,
     };
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -300,6 +300,9 @@ export class EventsService {
       _sum: {
         points: true,
       },
+      _count: {
+        points: true,
+      },
       where: {
         type,
         user_id: id,
@@ -307,13 +310,7 @@ export class EventsService {
       },
     });
     return {
-      count: await this.prisma.event.count({
-        where: {
-          type,
-          user_id: id,
-          deleted_at: null,
-        },
-      }),
+      count: aggregate._count.points || 0,
       points: aggregate._sum.points || 0,
     };
   }
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -136,7 +136,7 @@ export class EventsService {
     }
 
     const [nextRecords = [], previousRecords = []] = await Promise.all([
-      await this.prisma.event.findMany({
+      this.prisma.event.findMany({
         where: {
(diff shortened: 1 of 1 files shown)


NullSoldier:
Thanks for the PR @Skakruk, we'll review it today.

src/events/events.service.ts
```diff
@@ -57,20 +57,26 @@ export class EventsService {
...
     const orderBy = {
       occurred_at: Prisma.SortOrder.desc,
     };
-    const skip = cursor ? 1 : 0;
+    const skip = cursorId ? 1 : 0;
+    const fn = options.before ? 'gte' : 'lte';
     const where = {
       user_id: options.userId,
       deleted_at: null,
+      ...(cursorId
```
rohanjadvani, line 71: This is a good change, but I think it's worth separating this additional check to make it more readable. What do you think of the following?

```ts
const where: Prisma.EventWhereInput = {
  user_id: options.userId,
  deleted_at: null,
};
if (cursorId) {
  if (options.before) {
    where.id = { gte: cursorId };
  } else {
    where.id = { lte: cursorId };
  }
}
```

src/events/events.service.ts
```diff
@@ -128,20 +134,29 @@ export class EventsService {
...
-    });
-    const previousRecords = await this.prisma.event.findMany({
-      where,
-      orderBy,
-      cursor: { id: data[0].id },
-      skip: 1,
-      take: -1,
-    });
+
+    const [nextRecords = [], previousRecords = []] = await Promise.all([
```
rohanjadvani, line 138: Can we remove the `Promise.all`? It's probably low risk here, but it can lead to db starvation. These queries should be fast anyways.

src/events/events.service.ts
Skakruk: yeah, it's much easier to read.

src/events/events.service.ts
Skakruk: makes sense

src/events/events.service.ts
```diff
@@ -284,20 +300,17 @@ export class EventsService {
       _sum: {
         points: true,
       },
+      _count: {
+        points: true,
+      },
```
Skakruk, line 305: @rohanjadvani I've added changes to `/metrics` route, since the main bottleneck was also in EventsService

rohanjadvani (approved):

rohanjadvani:
Thanks @Skakruk - I'll get this deployed today.

PR #980, opened 2022-07-26
Use occurred_at field as cursor

Skakruk:
## Summary

Use `event`s `occurred_at` property as a cursor, since events are always ordered by `occurred_at`.
Previously used `id` as cursor didn't work well when an event was added with `occurred_at` in the past. Its `id` was greater than a more recent event. This led to fetching an incorrect list of events.

## Breaking Change

Needs the front-end part to be updated as well.  

- [ ] Yes

### DIFF
--- a/src/events/dto/events-query.dto.ts
+++ b/src/events/dto/events-query.dto.ts
@@ -3,14 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { ApiPropertyOptional } from '@nestjs/swagger';
 import { Type } from 'class-transformer';
-import {
-  IsInt,
-  IsISO8601,
-  IsNumber,
-  IsOptional,
-  Max,
-  Min,
-} from 'class-validator';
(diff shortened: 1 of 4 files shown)


Skakruk:
Oh, I didn't notice that it could be multiple records with equally the same timestamp (up to milliseconds). In this case, yes, those 30 records will be skipped.
However, currently also exist the problem, is when some records can be skipped.
let's say, we have the next records in DB, sorted by `occurred_at`

| id | occurred_at |
| --- | --- | 
| 1  | 2022-07-29T10:00:00.000Z
| 2  | 2022-07-29T09:00:00.000Z
| 5  | 2022-07-29T08:00:00.000Z
| 4  | 2022-07-29T07:00:00.000Z
| 3  | 2022-07-29T06:00:00.000Z
| 6  | 2022-07-29T05:00:00.000Z

When we query the first time without a cursor with limit 3 - we get records 1,2,5.
When we query the next batch with cursor > 5 - we get only a record with id 6.

Anyway, we need unique, sequential column, and data should be sorted by this column.

PR #982, opened 2022-07-28
Set dependabot to only update once a week

mat-if:
## Summary

Allows us to minimize time spent on looking at/merging updates a little bit

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):
🙏

PR #985, opened 2022-07-31
Add ability to skip measurements in telemetry

danield9tqh:
## Summary
We are running up against InfluxDB usage limits. There are some measurements we are not using currently but still want to keep in telemetry so we don't have to do another release if we remove them. This allows us to keep them in telemetry but not upload to influx

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/telemetry/telemetry.controller.ts
+++ b/src/telemetry/telemetry.controller.ts
@@ -52,6 +52,10 @@ export class TelemetryController {
       }
       nodeVersion = version.value;
 
+      if (this.getSkippedMeasurements().includes(measurement)) {
+        continue;
+      }
+
       options.push({
         fields,
         measurement,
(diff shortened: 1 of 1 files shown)


src/telemetry/telemetry.controller.ts
```diff
@@ -90,6 +94,14 @@ export class TelemetryController {
     }
   }
 
+  private getSkippedMeasurements(): string[] {
+    const measurements = this.config.getWithDefault<string>(
+      'SKIP_MEASUREMENTS',
+      '',
+    );
+    return measurements ? measurements.split(',') : [];
```
rohanjadvani, line 102: We don't need to check if `measurement` is undefined anymore, right?

danield9tqh: Not with the `getWithDefault` I don't think

danield9tqh: But this is there in case it returns `''`. `''.split(',')` will return `[""]` but I wanted it to just return `[]`. 🤷🏻 not that important but slightly more accurate

rohanjadvani (approved):

PR #991, opened 2022-08-03
Add telemetry whitelisting

NullSoldier:
## Summary

This whitelists all telemetry points. If you want telemetry in here,
you'll need to submit a code review after you submit it to the node
to let it through into our influx instance.

Likewise, we now only let every 5th block_propagation block sequence
point through.

## Testing Plan
Run new tests, and watch our influx dashboard.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/telemetry/telemetry.controller.ts
+++ b/src/telemetry/telemetry.controller.ts
@@ -26,7 +26,7 @@ import {
 /** How many blocks per sequence for block_propagation measurement to allow through telemetry */
 export const BLOCK_PROPAGATION_INTERVAL = 5;
 
-const TELEMETRY_WHITELIST = new Map<string, true | RegExp | Array<string>>([
+const TELEMETRY_WHITELIST = new Map<string, true | Array<string>>([
   ['block_mined', true],
   ['block_propagation', true],
   ['node_started', true],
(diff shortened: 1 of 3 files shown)


src/telemetry/telemetry.controller.ts
```diff
@@ -11,13 +11,41 @@ import {
...
-import { WriteTelemetryPointsDto } from './dto/write-telemetry-points.dto';
+import {
+  WriteTelemetryPointDto,
+  WriteTelemetryPointsDto,
+} from './dto/write-telemetry-points.dto';
+
+/** How many blocks per sequence for block_propagation measurement to allow through telemetry */
+export const BLOCK_PROPAGATION_INTERVAL = 5;
+
+const TELEMETRY_WHITELIST = new Map<string, true | RegExp | Array<string>>([
```
rohanjadvani, line 29: nit: I don't feel strongly about this, but if we're not using any `RegExp` now maybe we can remove it and add it in the future.

rohanjadvani (approved):
Looks good after lint is fixed

src/telemetry/telemetry.controller.ts
NullSoldier: I think it'll be needed by someone soon, but I generally agree with the premise of no code that is not planned to be used very soon. I'll remove it for now.

NullSoldier: Fixed

hughy (approved):

PR #993, opened 2022-08-05
Add network message types to telemetry whitelist

deekerno:
## Summary
Adds individual network message types to telemetry whitelist.

## Testing Plan
Run tests, and watch influx dashboard.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/telemetry/telemetry.controller.ts
+++ b/src/telemetry/telemetry.controller.ts
@@ -26,38 +26,6 @@ import {
 /** How many blocks per sequence for block_propagation measurement to allow through telemetry */
 export const BLOCK_PROPAGATION_INTERVAL = 5;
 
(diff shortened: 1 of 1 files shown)


src/telemetry/telemetry.controller.ts
```diff
@@ -37,11 +69,11 @@ const TELEMETRY_WHITELIST = new Map<string, true | Array<string>>([
       'heap_used',
       'peers_count',
       'session_id',
-      'inbound_traffic',
-      'outbound_traffic',
```
NullSoldier, line 41: Can we bring these back? They're very useful for monitoring network conditions.

deekerno: They should still be present by way of `...traffic_types` below. I can remove that line if it's less clear.

danield9tqh (approved):

PR #1053, opened 2022-10-21
feat: make env vars and endpoint for increasing testnet deposits

jowparks:
## Summary

I see no staging branch so I am pointing this to main.

This PR does 2 things:
1. Adds min an max deposit which will be used in the CLI to determine min and max deposit at runtime. PR for that is here:
https://github.com/iron-fish/ironfish/pull/2391
2. Creates an InfluxDB call that will occur every 10 minutes to emit a metric to InfluxDB
## Testing Plan
Used README to boot up locally

Ran GET against `localhost:8003/deposits/min_and_max_deposit`
Output:
```
{
    "minDeposit": "0.1",
    "maxDeposit": "1"
}
```
## Breaking Change

```
[ ] Yes
```


.env.test
```diff
@@ -15,6 +15,8 @@ INFLUXDB_ORG=Iron Fish
 INFLUXDB_URL=http://localhost:8004 
 IRONFISH_API_KEY=test
 MAGIC_SECRET_KEY=test
```
jowparks, line 17: What do we want to make this number?

src/events/deposits.upsert.service.ts
```diff
@@ -255,4 +257,37 @@ export class DepositsUpsertService {
       );
     });
   }
+
+  async syncDepositedIron(): Promise<void> {
+    const aggregate = await this.prisma.deposit.aggregate({
+      _sum: {
+        amount: true,
+      },
+    });
```
jowparks, line 266: This requires dependency injection of `influxDbService`, but I don't know where to add it. There is no corresponding `*.module.ts`

Note that I see this error when I run a unittest locally.

src/events/deposits.upsert.service.ts
```diff
@@ -255,4 +257,37 @@ export class DepositsUpsertService {
...
+        tags: [],
+        timestamp: new Date(),
+      },
+    ]);
+    const runAt = new Date();
+    runAt.setMinutes(runAt.getMinutes() + 10);
+    await this.graphileWorkerService.addJob(
+      GraphileWorkerPattern.SYNC_DEPOSITED_IRON,
+      {},
+      {
```
jowparks, line 286: Do I need to define this queue somewhere, or will it autocreate if it doesn't exist already?

src/events/deposits.upsert.service.ts
```diff
@@ -255,4 +257,37 @@ export class DepositsUpsertService {
...
   }
+
+  async syncDepositedIron(): Promise<void> {
+    const aggregate = await this.prisma.deposit.aggregate({
+      _sum: {
+        amount: true,
+      },
+    });
+    this.influxDbService.writePoints([
+      {
```
jowparks, line 268: For influxdb, will this `measurement` be autocreated as a category when the first measurement is sent? Or do I define somewhere?

.env.test
jowparks: cc @NullSoldier

src/events/deposits.upsert.service.ts
rohanjadvani: You don't need to explicitly define this anywhere.

src/events/deposits.upsert.service.ts
rohanjadvani: The queue name is a column value in the `graphile_worker` database, so you don't need to define it anywhere.

src/graphile-worker/enums/graphile-worker-pattern.ts
```diff
@@ -2,14 +2,15 @@
...
-  DELETE_BLOCK_MINED_EVENT = 'DELETE_BLOCK_MINED_EVENT',
   CREATE_NODE_UPTIME_EVENT = 'CREATE_NODE_UPTIME_EVENT',
-  REFRESH_USERS_POINTS = 'REFRESH_USERS_POINTS',
+  DELETE_BLOCK_MINED_EVENT = 'DELETE_BLOCK_MINED_EVENT',
+  REFRESH_DEPOSIT = 'REFRESH_DEPOSIT',
+  REFRESH_DEPOSITS = 'REFRESH_DEPOSITS',
   REFRESH_USER_POINTS = 'REFRESH_USER_POINTS',
+  REFRESH_USERS_POINTS = 'REFRESH_USERS_POINTS',
   SYNC_BLOCKS_DAILY = 'SYNC_BLOCKS_DAILY',
+  SYNC_DEPOSITED_IRON = 'SYNC_DEPOSITED_IRON',
```
rohanjadvani, line 12: What do you think about `SUBMIT_DEPOSITED_IRON_TO_TELEMETRY`?

src/events/deposits.controller.ts
```diff
@@ -80,4 +91,13 @@ export class DepositsController {
       GraphileWorkerPattern.REFRESH_DEPOSITS,
     );
   }
+
+  @ApiExcludeEndpoint()
+  @Post('deposited_iron')
```
rohanjadvani, line 96: What do you think about `'sync_to_telemetry'`?

src/events/deposits.controller.ts
```diff
@@ -55,6 +55,17 @@ export class DepositsController {
...
 
+  @ApiOperation({
+    summary:
+      'The min and max increment of IRON allowed to be deposited to Iron Bank',
+  })
+  @Get('min_and_max_deposit')
+  minAndMaxDeposit(): { minDeposit: number; maxDeposit: number } {
+    const maxDeposit = this.configService.get<number>('MAX_DEPOSIT');
+    const minDeposit = this.configService.get<number>('MIN_DEPOSIT');
+    return { minDeposit, maxDeposit };
```
rohanjadvani, line 66: We generally use snake case in returned fields from the API
```suggestion
    return { min_deposit, max_deposit };
```

rohanjadvani: What do you think about `min_deposit_size` / `max_deposit_size` or `min_size` / `max_size`?

src/events/deposits.controller.spec.ts
```diff
@@ -206,4 +206,21 @@ describe('DepositsController', () => {
       );
     });
   });
+
```
rohanjadvani, line 209: Can we add a test for the deposit range endpoint?

rohanjadvani (commented):
Some minor nits, but also would like to add a test case.

.env.test
jowparks: Going to set 0.6 per our discussion and we can ramp up if needed

src/events/deposits.upsert.service.ts
```diff
@@ -255,4 +257,37 @@ export class DepositsUpsertService {
...
+    ]);
+    const runAt = new Date();
+    runAt.setMinutes(runAt.getMinutes() + 10);
+    await this.graphileWorkerService.addJob(
+      GraphileWorkerPattern.SUBMIT_DEPOSITED_IRON_TO_TELEMETRY,
+      {},
+      {
+        jobKeyMode: `preserve_run_at`,
+        queueName: `deposited_iron`,
+        runAt,
```
jowparks, line 289: What is the suggested code change here?

src/events/deposits.controller.ts
```diff
@@ -80,4 +91,13 @@ export class DepositsController {
       GraphileWorkerPattern.REFRESH_DEPOSITS,
     );
   }
+
+  @ApiExcludeEndpoint()
+  @Post('sync_to_telemetry')
+  @UseGuards(ApiKeyGuard)
+  async depositedIron(): Promise<void> {
```
rohanjadvani, line 98:
```suggestion
  async syncToTelemetry(): Promise<void> {
```

src/events/deposits.jobs.controller.ts
```diff
@@ -30,6 +30,13 @@ export class DepositsJobsController {
     return { requeue: false };
   }
 
+  @MessagePattern(GraphileWorkerPattern.SUBMIT_DEPOSITED_IRON_TO_TELEMETRY)
+  @UseFilters(new GraphileWorkerException())
+  async syncDepositedIron(): Promise<GraphileWorkerHandlerResponse> {
```
rohanjadvani, line 35:
```suggestion
  async submitDepositedIronToTelemetry(): Promise<GraphileWorkerHandlerResponse> {
```

src/events/deposits.upsert.service.ts
```diff
@@ -255,4 +257,33 @@ export class DepositsUpsertService {
...
+        tags: [],
+        timestamp: new Date(),
+      },
+    ]);
+    const runAt = new Date();
+    runAt.setMinutes(runAt.getMinutes() + 10);
+    await this.graphileWorkerService.addJob(
+      GraphileWorkerPattern.SUBMIT_DEPOSITED_IRON_TO_TELEMETRY,
+      {},
+      { jobKey: 'deposit_sync_to_telemetry', runAt: runAt },
```
rohanjadvani, line 286:
```suggestion
      { jobKey: 'sync_deposited_iron_to_telemetry', runAt },
```

rohanjadvani (approved):
Some rename suggestions for consistency. Looks good otherwise

PR #1060, opened 2022-10-26
feat: add logic to adjust deposit points for increased deposit size

jowparks:
## Summary
Scales the amount of points based on deposit size. This is related to the PRs:
https://github.com/iron-fish/ironfish/pull/2391
https://github.com/iron-fish/ironfish-api/pull/1053

## Testing Plan
Unittests

### DIFF
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -457,7 +457,14 @@ export class EventsService {
       return null;
     }
 
-    const adjustedPoints = points ?? POINTS_PER_CATEGORY[type];
+    let adjustedPoints = points ?? POINTS_PER_CATEGORY[type];
+
+    if (deposit) {
+      adjustedPoints =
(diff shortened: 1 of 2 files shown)


src/events/events.service.ts
```diff
@@ -457,7 +458,16 @@ export class EventsService {
       return null;
     }
 
-    const adjustedPoints = points ?? POINTS_PER_CATEGORY[type];
+    let adjustedPoints = points ?? POINTS_PER_CATEGORY[type];
```
NullSoldier, line 461: This is small but we actually always need to take the points as an override because it's used by our manual control panel to create events with manually set points.

1. points
2. if deposit use manual calculation
3. otherwise use POINTS_PER_CATEGORY

NullSoldier (approved):

PR #1064, opened 2022-10-31
feat: add masp transactions data model and migrations

jowparks:
## Summary
Adds MASP transaction db model, migration, client codegen.

Open for debate if this is the best modeling path. I wasn't sure why `Event` is a separate model from `Deposit`, but I followed that pattern.

I also could imagine making these `Send`, `Mint`, and `Burn` separate models, wasn't sure which was preferable. 
## Breaking Change

```
[ ] Yes
```


prisma/migrations/20221021163739_drop_and_create_index/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- DropIndex
```
jowparks, line 1: This file was autogenerated for me, not sure why, we can delete if needed.

package.json
```diff
@@ -27,8 +27,8 @@
     "test:keepdb": "dotenv -e .env.test -- jest --runInBand",
     "test": "dotenv -e .env.test -- prisma migrate reset --force --skip-generate > /dev/null && yarn test:keepdb",
     "db:client:generate": "prisma generate",
-    "db:migrate": "prisma migrate dev",
-    "db:client:format": "prisma format"
+    "db:client:format": "prisma format",
```
rohanjadvani, line 30: Should `master` be rebased into this branch?

prisma/schema.prisma
```diff
@@ -229,3 +231,28 @@ model DepositHead {
...
+  block_hash       String              @db.VarChar
+  asset            String              @db.VarChar
+  type             MaspTransactionType
+  block_sequence   Int
+  network_version  Int
+  main             Boolean
+  amount           Int
+  event            Event?
+
+  eventId Int?
```
rohanjadvani, line 249:
```suggestion
  event_id Int?
```

rohanjadvani: Will these be unique?

prisma/schema.prisma
```diff
@@ -229,3 +231,28 @@ model DepositHead {
...
   @@map("deposit_head")
 }
+
+model MaspTransaction {
+  id               Int                 @id @default(autoincrement())
+  created_at       DateTime            @default(now()) @db.Timestamp(6)
+  updated_at       DateTime            @default(now()) @updatedAt @db.Timestamp(6)
+  transaction_hash String              @db.VarChar
+  block_hash       String              @db.VarChar
+  asset            String              @db.VarChar
```
rohanjadvani, line 241: We've been using asset identifier in the MASP prototypes. What do you think about that?

prisma/schema.prisma
```diff
@@ -229,3 +231,28 @@ model DepositHead {
...
+  amount           Int
+  event            Event?
+
+  eventId Int?
+
+  @@map("masp_transactions")
+}
+
+enum MaspTransactionType {
+  SEND
```
rohanjadvani, line 255: What do you think about `TRANSFER`? I know we had `SEND` in the spec, but we've been using that vocabulary in the MASP prototypes

package.json
jowparks: I already rebased, I think I flipped order is all (had created PR for format after this one).

prisma/schema.prisma
jowparks: I am not sure for these two questions, I was imagining the event represents a sent event from client. Not sure if we want to dedupe and prevent secondary save or if we want to save both and have one to many for masp <- event. I think you guys would be better at making the decision on that given you have been through a couple testnet phases already

prisma/schema.prisma
jowparks: SGTM

prisma/schema.prisma
```diff
@@ -229,3 +231,28 @@ model DepositHead {
...
   @@map("deposit_head")
 }
+
+model MaspTransaction {
+  id               Int                 @id @default(autoincrement())
+  created_at       DateTime            @default(now()) @db.Timestamp(6)
+  updated_at       DateTime            @default(now()) @updatedAt @db.Timestamp(6)
+  transaction_hash String              @db.VarChar
+  block_hash       String              @db.VarChar
+  asset_identifier String              @db.VarChar
```
jowparks, line 241: I was guessing this identifier is a hash, so I put as string. Wondering if we want any other attributes of the assets/transactions and if we want to index on the identifier.

prisma/schema.prisma
```diff
@@ -229,3 +231,28 @@ model DepositHead {
...
+  type             MaspTransactionType
+  block_sequence   Int
+  network_version  Int
+  main             Boolean
+  amount           Int
+  event            Event?
+
+  event_id Int?
```
rohanjadvani, line 249: I don't think this `event_id` is necessary. I imagine this should look fairly similar to `Deposit`

prisma/schema.prisma
rohanjadvani: I don't think we need to search MASP transactions based on a given asset yet, so I think we can omit for now.

prisma/migrations/20221031181536_add_masp_transactions/migration.sql
```diff
@@ -0,0 +1,35 @@
...
+ALTER TABLE "events" ADD COLUMN     "masp_transaction_id" INTEGER;
+
+-- CreateTable
+CREATE TABLE "masp_transactions" (
+    "id" SERIAL NOT NULL,
+    "created_at" TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated_at" TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "transaction_hash" VARCHAR NOT NULL,
+    "block_hash" VARCHAR NOT NULL,
+    "asset" VARCHAR NOT NULL,
```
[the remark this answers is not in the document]
jowparks: Yup you are right, deleted the migration and regenerated.

rohanjadvani (approved):

PR #1075, opened 2022-11-02
Use Ore to calculate point, instead of IRON

hairtail:
## Breaking Change

```
[ ] Yes
```


jowparks:
Nice find!

jowparks (approved):

PR #1078, opened 2022-11-02
feat(api): Create Graphile job for event creation

rohanjadvani:
## Summary

Retool is timing out with event creation. This code change enqueues a worker job so it can be picked up by Graphile later.

## Testing Plan

Added unit tests.

## Breaking Change

```
[ ] Yes
```


src/events/events.controller.ts
```diff
@@ -71,22 +76,21 @@ export class EventsController {
...
-      userId: user.id,
-      url,
-    });
-
-    if (!event) {
-      return null;
-    }
-
-    return serializedEventFromRecordWithMetadata(event);
+    await this.graphileWorkerService.addJob(
```
jowparks, line 83: Do we have any concern with the create event endpoint getting called very often? I know this was basically how the `deposit` jobs ran out of control. If this is a relatively infrequently used job, seems fine to me.

jowparks (approved):

src/events/events.controller.ts
rohanjadvani: This won't be used that often. It's used for manual events from Retool.

PR #1081, opened 2022-11-02
Remove per-transaction upserts and replace with bulk createMany

dguenther:
## Summary

Transactions are already unique on hash + network version, and transactions should never change once they've been created, so we can use the bulk createMany operation to create Transactions and BlockTransactions.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/transactions/transactions.service.ts
+++ b/src/transactions/transactions.service.ts
@@ -24,53 +24,48 @@ export class TransactionsService {
     private readonly prisma: PrismaService,
   ) {}
 
-  async bulkUpsertWithClient(
+  async createManyWithClient(
     prisma: BasePrismaClient,
     transactions: UpsertTransactionOptions[],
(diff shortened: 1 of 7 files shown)


danield9tqh (approved):

src/transactions/transactions.service.ts
```diff
@@ -24,56 +24,51 @@ export class TransactionsService {
...
+        network_version: networkVersion,
+        fee: tx.fee,
+        size: tx.size,
+        notes: classToPlain(tx.notes),
+        spends: classToPlain(tx.spends),
+      })),
+      skipDuplicates: true,
     });
+
+    return await this.findMany(prisma, transactions);
```
rohanjadvani, line 45:
```suggestion
    return this.findMany(prisma, transactions);
```

rohanjadvani (approved):

PR #1083, opened 2022-11-03
feat: further model updates to support masp

jowparks:
## Summary
Further model updates that are required for supporting masp transactions in the API
related to:
https://github.com/iron-fish/ironfish-api/pull/1064

## Breaking Change

```
[ ] Yes
```


prisma/schema.prisma
```diff
@@ -245,13 +258,15 @@ model MaspTransaction {
   transaction_hash String              @db.VarChar
   block_hash       String              @db.VarChar
   asset_identifier String              @db.VarChar
+  graffiti         String              @db.VarChar
```
jowparks, line 261: We should probably store graffiti on masp right?

prisma/schema.prisma
```diff
@@ -170,6 +171,9 @@ enum EventType {
   SOCIAL_MEDIA_PROMOTION
   NODE_UPTIME
   SEND_TRANSACTION
+  MASP_TRANSFER
+  MASP_BURN
+  MASP_MINT
```
jowparks, line 176: Enum previously didn't have entries for event type

prisma/schema.prisma
```diff
@@ -245,13 +258,15 @@ model MaspTransaction {
...
   asset_identifier String              @db.VarChar
+  graffiti         String              @db.VarChar
   type             MaspTransactionType
   block_sequence   Int
   network_version  Int
   main             Boolean
   amount           Int
   event            Event?
 
+  @@unique([transaction_hash, graffiti, type], name: "uq_masp_on_transaction_hash_graffiti_type", map: "uq_masp_on_transaction_hash_graffiti_type")
```
jowparks, line 269: Does this unique constraint seem correct?

prisma/schema.prisma
rohanjadvani: How are these events going to be categorized? Will there be a single MASP type? I'm wondering if we should have types here and remove the type from the `masp_transaction` table.

jowparks: Seems reasonable to me, I can remove from the `masp_transaction` table

prisma/schema.prisma
```diff
@@ -239,24 +252,25 @@ model DepositHead {
...
-  block_hash       String              @db.VarChar
-  asset_identifier String              @db.VarChar
-  type             MaspTransactionType
+  id               Int       @id @default(autoincrement())
+  created_at       DateTime  @default(now()) @db.Timestamp(6)
+  updated_at       DateTime  @default(now()) @updatedAt @db.Timestamp(6)
+  transaction_hash String    @db.VarChar
+  block_hash       String    @db.VarChar
+  asset_name       String    @db.VarChar
+  type             EventType
```
jowparks, line 261: Jason suggested we include the type here so that if the events get deleted we will still have the record. He also said just to use the `EventType` enum and we can only ever use the `MASP*` for these entries.

rohanjadvani (approved):

PR #1087, opened 2022-11-07
Optimizes DepsoitsUpsertService.upsert()

NullSoldier:
## Summary

Converts this code path to use many bulk creation strategies instead
of upserting one at a time in the block.

## Testing Plan

I pulled down production DB and ran it locally. Takes about 4s per 300 TX instead of 40 seconds.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/deposits.upsert.service.ts
+++ b/src/events/deposits.upsert.service.ts
@@ -257,6 +257,8 @@ export class DepositsUpsertService {
               amount: number;
             }>();
 
+            const graffitis = new Array<string>();
+
             for (const transaction of operation.transactions) {
               const amounts = new Map<string, number>();
 
@@ -278,13 +280,13 @@ export class DepositsUpsertService {
                   main: true,
                   network_version: networkVersion,
                 });
+
+                graffitis.push(graffiti);
               }
             }
 
             // Bulk load unique users and map by graffiti
-            users = await this.usersService.findManyAndMapByGraffiti(
-              depositParams.map((d) => d.graffiti),
-            );
+            users = await this.usersService.findManyAndMapByGraffiti(graffitis);
 
             // Filter deposits made by unknown users
             depositParams = depositParams.filter((d) => users.has(d.graffiti));
@@ -319,6 +321,7 @@ export class DepositsUpsertService {
 
             const eventPayloads = [];
             const usersFiltered = new Map<string, User>();
+            const eventDepositIds = new Array<number>();
 
             // Create the event payloads filtering events and users on points < 0
             for (const deposit of deposits) {
@@ -336,6 +339,7 @@ export class DepositsUpsertService {
               }
 
               usersFiltered.set(deposit.graffiti, user);
+              eventDepositIds.push(deposit.id);
 
               eventPayloads.push({
                 occurred_at: blockTimestamp.toISOString(),
@@ -352,7 +356,7 @@ export class DepositsUpsertService {
             await prisma.event.deleteMany({
               where: {
                 deposit_id: {
-                  in: deposits.map((deposit) => deposit.id),
+                  in: eventDepositIds,
                 },
               },
             });
--- a/src/events/deposits.upsert.service.ts
+++ b/src/events/deposits.upsert.service.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { Injectable } from '@nestjs/common';
-import { Deposit, EventType, Prisma } from '@prisma/client';
+import { Deposit, EventType, Prisma, User } from '@prisma/client';
 import is from '@sindresorhus/is';
 import assert from 'assert';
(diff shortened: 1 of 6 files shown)


src/events/deposits.upsert.service.ts
```diff
@@ -288,4 +229,153 @@ export class DepositsUpsertService {
...
+            }
+          }
+
+          // Bulk load unique users and map by graffiti
+          const users = await this.usersService.findManyAndMapByGraffiti(
+            Array.from(amounts.keys()),
+          );
+
+          // Create deposit params for each deposit with a matching user
+          const depositParams = new Array<{
```
jowparks, line 270: Probably can use a client generated type here, maybe `DepositUpdateInput`

src/events/deposits.upsert.service.ts
```diff
@@ -288,4 +229,153 @@ export class DepositsUpsertService {
...
+              AND: depositParams.map((deposit) => ({
+                transaction_hash: deposit.transaction_hash,
+                graffiti: deposit.graffiti,
+              })),
+              network_version: networkVersion,
+            },
+          });
+
+          // Now create new not existing deposits
+          await prisma.deposit.createMany({
```
jowparks, line 313: Weird that this doesn't return ids of created entries :/ Looked at source, only returns counts

NullSoldier: It's a postgres thing

src/events/deposits.upsert.service.ts
```diff
@@ -288,4 +229,153 @@ export class DepositsUpsertService {
...
+          });
+
+          deposits = await prisma.deposit.findMany({
+            where: {
+              block_hash: blockHash,
+              network_version: networkVersion,
+            },
+          });
+
+          const eventPayloads = deposits.map((deposit) => {
```
jowparks, line 325: Doubt that the number of loop iterations matter here, but a friend once told me about how bad `map` is in terms of performance vs `for` loop. Makes me sad since I like syntax/paradigm of `map` better.

https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/

src/events/deposits.upsert.service.ts
jowparks: Scratch this comment, I just tried importing and I get a type error due to some ugly codegening in the library, too bad.

src/events/deposits.upsert.service.ts
```diff
@@ -288,4 +229,153 @@ export class DepositsUpsertService {
...
+            return {
+              occurred_at: operation.block.timestamp.toISOString(),
+              type: EventType.SEND_TRANSACTION,
+              user_id: user.id,
+              points: this.eventsService.calculateDepositPoints(deposit),
+              deposit_id: deposit.id,
+            };
+          });
+
+          await prisma.event.createMany({
```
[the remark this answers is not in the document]
NullSoldier: Yes, we need to do bulk operations here and control the queries tightly in this service.

src/events/deposits.upsert.service.ts
```diff
@@ -288,4 +228,190 @@ export class DepositsUpsertService {
...
+              });
+            }
+
+            users = usersFiltered;
+
+            // This SHOULD NOT be needed but this cleans up bad data in the server and crashes without this
+            await prisma.event.deleteMany({
+              where: {
+                deposit_id: {
+                  in: deposits.map((deposit) => deposit.id),
```
hughy, line 355: nit - we can gather deposit IDs in the loop above instead of iterating again

src/events/deposits.upsert.service.ts
```diff
@@ -288,4 +228,190 @@ export class DepositsUpsertService {
...
+                },
+              },
+            });
+
+            users = await this.usersService.findManyAndMapByGraffiti(
+              deposits.map((d) => d.graffiti),
```
hughy, line 390: nit - two loops over deposits here and on 384, above

hughy (approved):

src/events/deposits.upsert.service.ts
NullSoldier: changed

src/events/deposits.upsert.service.ts
```diff
@@ -288,4 +228,190 @@ export class DepositsUpsertService {
...
+                  block_sequence: operation.block.sequence,
+                  main: true,
+                  network_version: networkVersion,
+                });
+              }
+            }
+
+            // Bulk load unique users and map by graffiti
+            users = await this.usersService.findManyAndMapByGraffiti(
+              depositParams.map((d) => d.graffiti),
```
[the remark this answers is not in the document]
NullSoldier: changed

src/events/deposits.upsert.service.ts
NullSoldier: I don't see that, do you mean the one in the connected branch?

hughy: The one up above in `deleteMany` to map over deposits to get the IDs

PR #1092, opened 2022-11-09
fix: node uptime batching

jowparks:
## Summary
Batches node uptime events into single batch job queue. WIP need to add tests.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/node-uptimes/node-uptimes-loader.ts
+++ b/src/node-uptimes/node-uptimes-loader.ts
@@ -22,7 +22,7 @@ export class NodeUptimesLoader {
     private readonly prisma: PrismaService,
   ) {}
 
-  async createEvent(props: CreateEventsProps): Promise<void> {
+  async createEvents(props: CreateEventsProps): Promise<void> {
     for (const { user, occurredAt } of props) {
       const uptime = await this.nodeUptimesService.get(user);
       if (!uptime || uptime.total_hours < NODE_UPTIME_CREDIT_HOURS) {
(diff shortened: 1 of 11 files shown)


src/node-uptimes/node-uptimes-loader.ts
```diff
@@ -16,26 +22,28 @@ export class NodeUptimesLoader {
...
+      if (!uptime || uptime.total_hours < NODE_UPTIME_CREDIT_HOURS) {
+        return;
       }
 
-      await this.nodeUptimesService.decrementCountedHoursWithClient(
-        uptime,
-        prisma,
-      );
-    });
+      await this.prisma.$transaction(async (prisma) => {
```
jowparks, line 32: We could optimize these functions to do bulk update, but I left this way now to make sure direction seems sound.

src/node-uptimes/node-uptimes.jobs.controller.ts
```diff
@@ -4,32 +4,60 @@
...
-    userId,
-    occurredAt,
-  }: CreateNodeUptimeEventOptions): Promise<GraphileWorkerHandlerResponse> {
-    const user = await this.usersService.find(userId);
-    if (!user) {
-      this.loggerService.error(`No user found for '${userId}'`, '');
-      return { requeue: false };
+  async createNodeUptimeEvents(
+    uptimeEvents: CreateNodeUptimeEventOptionsList,
+  ): Promise<GraphileWorkerHandlerResponse> {
```
jowparks, line 26: dedupes here

PR #1100, opened 2022-11-11
Optimize updateLatestPoints

NullSoldier:
## Summary

This turns 3 queries into 1

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -606,7 +606,6 @@ export class EventsService {
     >(
       `SELECT
         SUM(points) AS total_points,
-        COUNT(*) AS total_count,
         SUM(CASE WHEN type = $1::event_type THEN points END) points,
         COUNT(CASE WHEN type = $1::event_type THEN 1 END) AS count,
         MAX(CASE WHEN type = $1::event_type THEN occurred_at END) AS last_occurred_at
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -604,14 +604,14 @@ export class EventsService {
         last_occurred_at: Date;
       }[]
     >(
-      `select
-        max(occurred_at) as last_occurred_at,
-        sum(points) as total_points,
-        count(*) as total_count,
-        sum(case when type = $1::event_type then points end) points,
-        count(case when type = $1::event_type then 0 end) as count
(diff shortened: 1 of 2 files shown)


src/events/events.service.ts
```diff
@@ -596,46 +596,36 @@ export class EventsService {
...
+    const [counts] = await this.prisma.readClient.$queryRawUnsafe<
+      {
+        count: BigInt;
+        points: BigInt;
+        total_points: BigInt;
+        last_occurred_at: Date;
+      }[]
+    >(
+      `select
+        max(occurred_at) as last_occurred_at,
```
NullSoldier, line 608: this needs to be filtered on the type

src/events/events.service.ts
```diff
@@ -596,46 +596,36 @@ export class EventsService {
...
+        sum(points) as total_points,
+        count(*) as total_count,
+        sum(case when type = $1::event_type then points end) points,
+        count(case when type = $1::event_type then 0 end) as count
```
NullSoldier, line 612: then 1

src/events/events.service.ts
NullSoldier: fixed

src/events/events.service.ts
NullSoldier: fixed

src/events/events.service.ts
```diff
@@ -596,46 +596,36 @@ export class EventsService {
...
-        points: true,
-      },
-      where: {
-        user_id: userId,
-      },
-    });
-    const totalPoints = totalPointsAggregate._sum.points ?? 0;
+    const [counts] = await this.prisma.readClient.$queryRawUnsafe<
+      {
+        count: BigInt;
```
jowparks, line 601: Looks like you are only using 4 of the 5 selected values? Also I take it ordering doesn't matter?

jowparks (approved):

src/events/events.service.ts
NullSoldier: yeah, order of keys in JS doesn't matter.

NullSoldier: Removed total_count.

PR #1103, opened 2022-11-11
Add check that deposit connection is correct

NullSoldier:
## Summary

Checks that a deposit operation is being applied cleanly, or fails. It requires fixing almost every test. You should review with https://github.com/iron-fish/ironfish-api/pull/1103/files?diff=split&w=1

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/deposits.upsert.service.ts
+++ b/src/events/deposits.upsert.service.ts
@@ -16,6 +16,7 @@ import { GraphileWorkerService } from '../graphile-worker/graphile-worker.servic
 import { InfluxDbService } from '../influxdb/influxdb.service';
 import { PrismaService } from '../prisma/prisma.service';
 import { UsersService } from '../users/users.service';
+import { DepositsService } from './deposits.service';
 import { UpsertDepositsOperationDto } from './dto/upsert-deposit.dto';
 import { EventsService } from './events.service';
 
@@ -24,6 +25,7 @@ export class DepositsUpsertService {
   constructor(
     private readonly config: ApiConfigService,
     private readonly depositHeadsService: DepositHeadsService,
+    private readonly depositsService: DepositsService,
     private readonly eventsService: EventsService,
     private readonly graphileWorkerService: GraphileWorkerService,
     private readonly influxDbService: InfluxDbService,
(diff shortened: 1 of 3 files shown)


src/events/deposits.upsert.service.ts
```diff
@@ -230,172 +232,194 @@ export class DepositsUpsertService {
...
+              deposit_id: {
+                in: eventDepositIds,
               },
-            });
+            },
+          });
+
+          await prisma.event.createMany({
+            data: eventPayloads,
+          });
```
jowparks, line 385: Wow that was a mega conditional

src/events/deposits.upsert.service.ts
```diff
@@ -230,172 +232,194 @@ export class DepositsUpsertService {
...
-                network_version: networkVersion,
-              },
-            });
+          users = usersFiltered;
 
-            deposits = await prisma.deposit.findMany({
-              where: {
-                block_hash: blockHash,
-                network_version: networkVersion,
+          // This SHOULD NOT be needed but this cleans up bad data in the server and crashes without this
```
jowparks, line 374: 👍

jowparks: Do we care if we lose history of updates here? This would delete the events that were marked `main: false` due to block disconnection, then it would recreate. Seems like the main risk is losing the `created_at` info for previously recorded events, but doesn't seem like a big deal as long as you are fine with it.

jowparks (approved):
Seems fine overall, just a couple concerns on some nuances.

src/events/deposits.upsert.service.ts
```diff
@@ -230,172 +232,194 @@ export class DepositsUpsertService {
...
-            for (const transaction of operation.transactions) {
-              const amounts = new Map<string, number>();
-
-              for (const note of transaction.notes) {
-                if (note.memo) {
-                  const amount = amounts.get(note.memo) ?? 0;
-                  amounts.set(note.memo, amount + note.amount);
-                }
-              }
+        if (operation.type === BlockOperation.CONNECTED) {
```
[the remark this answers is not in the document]
NullSoldier: The way the codebase is set up, you always want to standardize them before using them to be safe. The the api validators don't do this. We could get rid of these usages if we just had the DTO objects do this.

src/events/deposits.upsert.service.ts
NullSoldier: yeah, that code also hasn't changed it's been in for a week or two in the optimize deposits code. The indent just changed here.

PR #1109, opened 2022-11-14
feat: exclude users from rankings that have 0 points

jowparks:
## Summary
1. Excludes users from rank list if they have 0 points
2. When user is searched for, return users even if they have no points
3. Lifetime points should return 0 for user (without rank query) if user has no points.

## Testing Plan
Unittest

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -696,22 +696,6 @@ export class EventsService {
     count: number | null;
     rank: number | null;
   }> {
-    // return early if user doesn't have any points
-    const userPoints = await this.userPointsService.findOrThrow(user.id);
-    const filteredPoints = events
-      .map(
-        (e) =>
-          userPoints[(e.toLowerCase() + '_points') as keyof typeof userPoints],
-      )
-      .reduce((sum, current) => Number(sum) + Number(current), 0);
-    if (filteredPoints === 0) {
-      return {
-        rank: null,
-        points: null,
-        count: null,
-      };
-    }
-
     const queryPoints = events
       .map((e) => e.toLowerCase() + '_points')
       .join(' + ');
@@ -767,13 +751,17 @@ export class EventsService {
     >(query, user.id);
 
     if (rank.length === 0) {
-      throw new Error(`User ${user.id} has no user_points entry`);
+      return {
+        rank: null,
+        points: null,
+        count: null,
+      };
     }
 
     return {
-      rank: rank[0].rank || null,
-      points: rank[0].points || null,
-      count: rank[0].count || null,
+      rank: rank[0].rank,
+      points: rank[0].points,
+      count: rank[0].count,
     };
   }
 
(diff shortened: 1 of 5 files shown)


src/users/users.service.ts
```diff
@@ -269,7 +271,7 @@ export class UsersService {
             ) AS rank
           FROM
             users
-          JOIN
+          INNER JOIN
```
rohanjadvani, line 274: If I inner join with `user_latest_events`, will the search show a user if they don't have points? That'd seem a little confusing to me if I want to look at my profile but haven't participated yet.

jowparks: Yup you are right, users would be filtered out here if they have no points. So should we only filter them from list result?

src/events/events.service.ts
```diff
@@ -702,6 +702,15 @@ export class EventsService {
       .map((e) => e.toLowerCase() + '_last_occurred_at')
       .join(', ');
 
+    // return early if user doesn't have any points
+    const userPoints = await this.userPointsService.findOrThrow(user.id);
+    if (userPoints.total_points === 0) {
+      return {
+        rank: 0,
```
NullSoldier, line 709: We should return null for the rank here

src/events/events.service.ts
```diff
@@ -691,7 +691,27 @@ export class EventsService {
...
     events: EventType[],
-  ): Promise<{ points: number; count: number; rank: number }> {
+  ): Promise<{
+    points: number | null;
+    count: number | null;
+    rank: number | null;
+  }> {
+    // return early if user doesn't have any points
+    const userPoints = await this.userPointsService.findOrThrow(user.id);
+    const filteredPoints = events
```
jowparks, line 701: yuck, but has to be done this way based on duck typing

src/events/events.service.ts
```diff
@@ -747,9 +771,9 @@ export class EventsService {
     }
 
     return {
-      rank: rank[0].rank,
-      points: rank[0].points,
-      count: rank[0].count,
+      rank: rank[0].rank || null,
+      points: rank[0].points || null,
+      count: rank[0].count || null,
```
NullSoldier, line 776: Remove these || null

src/users/users.controller.spec.ts
```diff
@@ -349,6 +349,40 @@ describe('UsersController', () => {
           },
         });
       });
+      it('returns 0 for user without any points', async () => {
```
NullSoldier, line 352: Rename this to returns null

src/users/users.controller.spec.ts
```diff
@@ -349,6 +349,40 @@ describe('UsersController', () => {
...
+          });
+
+        expect(body).toMatchObject({
+          user_id: user.id,
+          granularity: MetricsGranularity.LIFETIME,
+          points: 0,
+          node_uptime: {
+            total_hours: 0,
+            last_checked_in: null,
+          },
```
NullSoldier, line 371: you could delete this since it is outside the scope of the test

NullSoldier:
Then we'll also need to upgrade getRank() as well.

NullSoldier (approved):

PR #1111, opened 2022-11-18
chore: update user metrics dto

jowparks:
## Summary
Realized we needed this change when I was trying to change the user profile page.

## Breaking Change

```
[ ] Yes
```


docker-compose.yml
```diff
@@ -11,6 +11,8 @@ services:
     volumes:
       - postgres_data:/var/lib/postgresql/data
       - ./docker-compose.sql:/docker-entrypoint-initdb.d/init.sql
```
jowparks, line 13: Put this in here for the next person wanting to do this

dguenther (approved):

PR #1115, opened 2022-11-21
chore: add assert to check empty transactions

jowparks:
## Summary
Simple assert

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/masp.upsert.service.ts
+++ b/src/events/masp.upsert.service.ts
@@ -52,6 +52,13 @@ export class MaspTransactionsUpsertService {
       'FORK operations not supported',
     );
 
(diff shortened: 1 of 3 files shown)


NullSoldier (approved):

src/events/masp.upsert.service.ts
```diff
@@ -191,8 +197,11 @@ export class MaspTransactionsUpsertService {
       return maspTransactions;
     });
```
jowparks, line 199: Had to add additional query here for disconnecting operations, when transactions aren't sent with disconnecting block, we can no long use the users calculated from `operation`.

PR #1123, opened 2022-11-23
Disable node hosting points

NullSoldier:
## Summary

We're turning off points for this to end phase 2.

## Breaking Change

```
[ ] Yes
```


hughy (approved):
Will we manually delete points that got in past the deadline?

src/telemetry/telemetry.controller.ts
```diff
@@ -105,7 +105,11 @@ export class TelemetryController {
   ): Promise<void> {
     const { options, nodeVersion } = this.processPoints(points);
 
-    if (graffiti && nodeVersion) {
+    if (
+      graffiti &&
+      nodeVersion &&
+      this.config.get<boolean>('ALLOW_NODE_UPTIME_POINTS')
```
hughy, line 111: nit - could we also bypass the call to `processPoints` on this condition?

PR #1143, opened 2022-12-09
Add leaderboard filters by pool

NullSoldier:
## Breaking Change

```
[ ] Yes
```


src/events/events.service.ts
```diff
@@ -237,6 +239,40 @@ export class EventsService {
...
       maspTransferAggregate.points;
 
+    const pool4Aggregates = [
+      bugCaughtAggregate,
+      nodeUptimeAggregate,
+      maspMintAggregate,
+      maspBurnAggregate,
+      maspTransferAggregate,
+    ];
+
```
jowparks, line 249: Bug?

src/user-points/user-points.jobs.controller.ts
```diff
@@ -68,4 +71,55 @@ export class UserPointsJobsController {
...
   }
+
+  @MessagePattern(GraphileWorkerPattern.REFRESH_POOL_4_POINTS)
+  @UseFilters(new GraphileWorkerException())
+  async refreshPool4Points({
+    userId,
+    endDate,
+  }: RefreshPool4Options): Promise<GraphileWorkerHandlerResponse> {
+    const user = await this.usersService.find(userId);
+    const end = endDate ?? new Date();
```
jowparks, line 82: This is purely here for testing, standard codepath will always land with `new Date()`

src/events/events.service.ts
NullSoldier: Those records were deleted

jowparks (approved):

PR #1144, opened 2022-12-09
Remove rank from users/find and users/[id]

NullSoldier:
## Summary

This was removed in the front end

Review with https://github.com/iron-fish/ironfish-api/pull/1144/files?diff=split&w=1

## Breaking Change

```
[ ] Yes
```


jowparks (approved):
Looks good, @NullSoldier removed the usage of rank on profile page for total points in this pr:
https://github.com/iron-fish/website-testnet/pull/340/files

PR #1156, opened 2022-12-20
chore: (2/2) multi asset naming cleanup

jowparks:
## Summary
Cleanup follow up after: https://github.com/iron-fish/ironfish-api/pull/1157
## Testing Plan
\## Breaking Change

```
[ ] Yes
```


prisma/schema.prisma
```diff
@@ -271,13 +271,13 @@ model MaspTransaction {
   main             Boolean
   event            Event?
```
jowparks, line 273: This new unique makes sense, right?

src/events/multi-asset.upsert.service.spec.ts
```diff
@@ -167,6 +167,9 @@ describe('MultiAssetUpsertService', () => {
     it('updates multiasset block hash on reorg', async () => {
       // setup
       await prismaService.multiAssetHead.deleteMany();
```
jowparks, line 169: I was seeing some leakage of test fixtures between these tests

jowparks:
no longer needed, wrapped everything in previous PR

PR #1163, opened 2023-01-05
fix: prevents retool migration issue on release

jowparks:
## Summary
Was running into error with deployment due to `retool` schema, to prevent this issue, we can drop and recreate schema on each deploy. This schema only contains views for internal usage.

For discussion surrounding issue see:
https://iron-fish.slack.com/archives/C02PF56HESK/p1672958477185669
## Breaking Change

```
[ ] Yes
```


rohanjadvani:
Discussed offline, but I think this can be deferred till post phase 3. Additionally it'd be nice if we moved this into a single script for local verification

rohanjadvani:
@jowparks I'm going to close this as per our Slack discussion. I think it's fine deferring to post Phase 3 launch, but give me a ping if you'd like to leave this open and continue work.

PR #1176, opened 2023-01-13
fix: link multiasset to blocks

jowparks:
## Summary

Enriching block hash and transaction hash directly from `multi_asset` model.
## Testing Plan
Screen Shot 2023-01-13 at 10 38 38 AM

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/events.service.ts
+++ b/src/events/events.service.ts
@@ -125,6 +125,16 @@ export class EventsService {
           block_hash: deposit.block_hash,
         };
       }
+      if (record.multi_asset_id) {
+        const multi_asset = await this.prisma.multiAsset.findFirstOrThrow({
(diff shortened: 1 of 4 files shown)


src/events/events.service.ts
```diff
@@ -125,6 +125,16 @@ export class EventsService {
           block_hash: deposit.block_hash,
         };
       }
+      if (record.multi_asset_id) {
+        const multi_asset = await this.prisma.multiAsset.findFirstOrThrow({
```
rohanjadvani, line 129: Can this use the MultiAssetService to encapsulate db interactions?

rohanjadvani (approved):
We should add some unit tests for the new service. Looks good otherwise

PR #1182, opened 2023-01-17
chore: add phase3 launch timestamp gate

jowparks:
## Summary
Gate phase 3 block uploads by timestamp on block submitted from syncer (which is a timestamp https://github.com/iron-fish/ironfish/blob/ea718ebc0145f8aef1ecc2214782f8eeadbcbd9c/ironfish/src/rpc/routes/chain/getTransactionStream.ts#L204)

Already added env vars to heroku for api as `PHASE3_LAUNCH_TIMESTAMP=1674064800000`
## Testing Plan
Unittests
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/events/multi-asset.upsert.service.ts
+++ b/src/events/multi-asset.upsert.service.ts
@@ -35,10 +35,19 @@ export class MultiAssetUpsertService {
         operation.type === BlockOperation.CONNECTED ||
         operation.type === BlockOperation.DISCONNECTED;
 
-      if (shouldUpsertMultiAsset) {
+      if (
+        shouldUpsertMultiAsset &&
+        this.occurredAfterLaunch(operation.block.timestamp)
+      ) {
         await this.upsert(operation);
       }
     }
+  }
+
+  private occurredAfterLaunch(comparisonDate: Date): boolean {
+    const timestamp = this.config.get<string>('PHASE3_LAUNCH_TIMESTAMP');
+    const launch = new Date(+timestamp);
+    return launch < comparisonDate;
   }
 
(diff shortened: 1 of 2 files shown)


src/events/multi-asset.upsert.service.ts
```diff
@@ -35,12 +35,21 @@ export class MultiAssetUpsertService {
...
+      if (
+        shouldUpsertMultiAsset &&
+        this.occurredAfterLaunch(operation.block.timestamp)
+      ) {
         await this.upsert(operation);
       }
     }
   }
 
+  occurredAfterLaunch = (comparisonDate: Date): boolean => {
```
rohanjadvani, line 47: nit for consistency with the rest of the code
```suggestion
  private occurredAfterLaunch(comparisonDate: Date): boolean {
```

src/events/multi-asset.upsert.service.ts
```diff
@@ -35,12 +35,21 @@ export class MultiAssetUpsertService {
...
+    const timestamp = this.config.get<string>('PHASE3_LAUNCH_TIMESTAMP');
+    const launch = new Date(+timestamp);
```
rohanjadvani, line 49: Can we just use `.get<number>`? Would that work?

jowparks: I put it as that previously, but it actually returns a string, so I changed it to this to match actual type. Was getting an invalid date because the value was actually a string (even though we told type system it was returning a number).

rohanjadvani (approved):
Looks good after tests pass. In the past we've just hard coded launch dates since they won't need to change, but I'll leave that decision up to you. We can leave as is.

jowparks:
No longer needed since launch has already happened

PR #1191, opened 2023-01-23
feat(api): Move uptime logic onto Graphile

rohanjadvani:
## Summary

* Enqueue graphile job every time on telemetry uptime 
* Check for hours in graphile job
* Remove advisory lock since all jobs run in a queue
* Refactor method name and tests

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```


rohanjadvani:
Close for now, seems like we're okay without it for now.

PR #1193, opened 2023-01-26
chore: deposit rescan

jowparks:
## Summary
This database change will be applied to `phase2`. There were some users who sent deposits to an incorrect address given by our company on twitter.

I will use this code locally, the payload I will use is below. I will need env var set for `MAX_DEPOSIT_SIZE=100000000` and `NETWORK_VERSION=3`, and `API_KEY=<key goes here>`, `DATABASE_URL=`

## Breaking Change

```
[ ] Yes
```


jowparks:
This was never intended to be merged. Closing to cleanup

PR #1228, opened 2023-02-14
feat(api): Create assets module

rohanjadvani:
## Summary

Create service to interact with assets table. Enables upserts, reads, and supply adjusting

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/assets/assets.service.ts
+++ b/src/assets/assets.service.ts
@@ -2,15 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { Injectable, NotFoundException } from '@nestjs/common';
-import { PrismaService } from '../prisma/prisma.service';
 import { BasePrismaClient } from '../prisma/types/base-prisma-client';
 import { CreateAssetOptions } from './interfaces/create-asset-options';
 import { Asset, Transaction } from '.prisma/client';
 
 @Injectable()
 export class AssetsService {
-  constructor(private readonly prisma: PrismaService) {}
-
   async findByIdentifierOrThrow(
     identifier: string,
     prisma: BasePrismaClient,
--- a/src/assets/assets.service.ts
+++ b/src/assets/assets.service.ts
@@ -1,0 +1,64 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Injectable, NotFoundException } from '@nestjs/common';
+import { PrismaService } from '../prisma/prisma.service';
+import { BasePrismaClient } from '../prisma/types/base-prisma-client';
+import { CreateAssetOptions } from './interfaces/create-asset-options';
+import { Asset, Transaction } from '.prisma/client';
+
+@Injectable()
+export class AssetsService {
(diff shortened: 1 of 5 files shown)


jowparks:
I think we are missing the model/client changes

rohanjadvani:
@jowparks I don't follow; what do you mean? Are you referring to the [db models](https://github.com/iron-fish/ironfish-api/blob/master/prisma/schema.prisma#L291-L329) or something else?

jowparks:
> @jowparks I don't follow; what do you mean? Are you referring to the [db models](https://github.com/iron-fish/ironfish-api/blob/master/prisma/schema.prisma#L291-L329) or something else?

Oh my bad I didn't have latest `master` and didn't see these models in my schema file

src/assets/assets.service.ts
```diff
@@ -0,0 +1,64 @@
...
+import { CreateAssetOptions } from './interfaces/create-asset-options';
+import { Asset, Transaction } from '.prisma/client';
+
+@Injectable()
+export class AssetsService {
+  constructor(private readonly prisma: PrismaService) {}
+
+  async findByIdentifierOrThrow(
+    identifier: string,
+    prisma: BasePrismaClient,
```
jowparks, line 16: remove this, the prisma service is already present via injection

rohanjadvani: I'm going to leave this for now, but remove the injection. The loader that I'm to add will pass this in for a database transaction. I will re-add the injection for the REST module later

src/assets/assets.service.ts
```diff
@@ -0,0 +1,61 @@
...
+    return prisma.asset.upsert({
+      create: {
+        identifier: options.identifier,
+        metadata: options.metadata,
+        name: options.name,
+        owner: options.owner,
+        supply: BigInt(0),
+        created_transaction_id: transaction.id,
+      },
+      update: {},
```
jowparks, line 40: I am not familiar with how the `update` vs `where` clause works with `prisma`. Can you update test to make sure that this doesn't update all assets (only one with provided `identifier`)?

src/assets/assets.service.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  async updateSupply(
+    asset: Asset,
+    delta: bigint,
+    prisma: BasePrismaClient,
+  ): Promise<Asset> {
+    return prisma.asset.update({
+      data: {
+        supply: asset.supply + delta,
```
jowparks, line 54: Will we run into any edge cases here if multiple `updateSupply` are happening at once? We could potentially revert the first delta right? I wonder if it is better to get supply and update in single transaction

rohanjadvani: This will all happen within a single transaction, which is why the `BasePrismaClient` is passed around in all these methods. They will be used from the loader in a subsequent PR.

src/assets/assets.service.ts
rohanjadvani: Prisma has a unique set of args based on the schema. For assets, it is:

```ts
  export type AssetWhereUniqueInput = {
    id?: number
    identifier?: string
  }
```

Since we have a unique constraint on identifier, I think this is pretty safe given the ORM docs and how we test `upsert` throughout the rest of the API. If you feel strongly though, I can update the test.

jowparks: All good, I just wasn't sure, thanks for context

jowparks (approved):

PR #1229, opened 2023-02-14
feat(api): Add asset descriptions module

rohanjadvani:
## Summary

Add module for asset descriptions

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/asset-descriptions/asset-descriptions.service.ts
+++ b/src/asset-descriptions/asset-descriptions.service.ts
@@ -1,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Injectable } from '@nestjs/common';
+import {
+  Asset,
+  AssetDescription,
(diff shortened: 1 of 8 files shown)


src/asset-descriptions/asset-descriptions.service.ts
```diff
@@ -0,0 +1,31 @@
...
+import { BasePrismaClient } from '../prisma/types/base-prisma-client';
+
+@Injectable()
+export class AssetDescriptionsService {
+  async create(
+    type: AssetDescriptionType,
+    value: bigint,
+    asset: Asset,
+    transaction: Transaction,
+    prisma: BasePrismaClient,
```
jowparks, line 20: Why use the `BasePrismaClient` here and not the `PrismaService`?

jowparks (approved):
Looks good, I think we want to inject the `PrismaService`, but other than that this one looks good

src/asset-descriptions/asset-descriptions.service.ts
rohanjadvani: I'm going to add a loader in a subsequent PR which will pass this argument from a database transaction. We won't need that service just yet, but will later. I'll add it when we need it.

PR #1230, opened 2023-02-14
feat(api): Add asset loader

rohanjadvani:
## Summary

Add module which will upsert assets and create/delete descriptions

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/assets-loader/assets-loader.ts
+++ b/src/assets-loader/assets-loader.ts
@@ -46,17 +46,7 @@ export class AssetsLoader {
     transaction: Transaction,
     prisma: BasePrismaClient,
   ): Promise<void> {
-    const existingDescriptions =
-      await this.assetDescriptionsService.findByTransaction(
-        transaction,
-        prisma,
-      );
-    if (existingDescriptions.length > 0) {
-      this.logger.debug(
-        `Skipping description loading for transaction '${transaction.hash}'`,
-      );
-      return;
-    }
+    await this.deleteAssetDescriptions(transaction, prisma);
 
     for (const mint of dto.mints) {
       const asset = await this.assetsService.upsert(
@@ -105,6 +95,14 @@ export class AssetsLoader {
         transaction,
         prisma,
       );
+
+    if (assetDescriptions.length === 0) {
+      return;
+    }
+
+    this.logger.debug(
+      `Deleting and re-processing descriptions for '${transaction.hash}'`,
+    );
 
     for (const assetDescription of assetDescriptions) {
       const asset = await this.assetsService.findOrThrow(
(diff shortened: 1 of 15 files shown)


src/assets-loader/assets-loader.ts
```diff
@@ -0,0 +1,127 @@
...
+    private readonly logger: LoggerService,
+    private readonly transactionsService: TransactionsService,
+  ) {}
+
+  async loadDescriptions(
+    block: BlockDto,
+    prisma: BasePrismaClient,
+  ): Promise<void> {
+    const main = block.type === BlockOperation.CONNECTED;
+
```
jowparks, line 29: Didn't you say the loader was going to use the PrismaService? Or is that happening in a subsequent PR:
https://github.com/iron-fish/ironfish-api/pull/1228#discussion_r1106348002

Just wondering because the replay of these might be tricky if it fails in the middle of updating a batch

src/assets-loader/assets-loader.ts
```diff
@@ -0,0 +1,127 @@
...
+        {
+          identifier: mint.id,
+          metadata: mint.metadata,
+          name: mint.name,
+          owner: mint.owner,
+        },
+        transaction,
+        prisma,
+      );
+
```
jowparks, line 72: What if we replay these, would there be new descriptions created? Should we delete the existing entries for a `transaction` before we create the new ones? We would also need to retroactively update the supply for the duplicates right?

Reason I ask is because it turned out to be very useful for `multi-asset` stuff to just replay the syncer from the beginning of time idempotently.

src/assets-loader/assets-loader.ts
rohanjadvani: This loader is using the prisma client from the block loader so mints and burns will be processed alongside transactions in the same transaction. I should have included more detail in my comment about where the original service was being injected from.

If these get slow, I'll move it into a worker job and then add a Prisma injection.

src/assets-loader/assets-loader.ts
rohanjadvani: I am assuming that if descriptions exist for a transaction hash, I'll skip processing [here](https://github.com/iron-fish/ironfish-api/pull/1230/files#diff-92a486d7d498d7daa503db784be52172c0e67eed3f0868e3e3bbf54271458afaR49-R59). I can change this to delete and then re-create. What do you think?

jowparks: Yea I think changing to delete/recreate is safer for replay, wouldn't require any manual modifications to DB.

jowparks (approved):
LGTM with changes we discussed

PR #1232, opened 2023-02-15
feat(api): Add `/assets/find`

rohanjadvani:
## Summary

Add endpoint to fetch asset by identifier

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/assets/assets.service.ts
+++ b/src/assets/assets.service.ts
@@ -24,15 +24,7 @@ export class AssetsService {
   }
 
   async findByIdentifierOrThrow(identifier: string): Promise<Asset> {
-    const record = await this.prisma.asset.findUnique({
-      where: {
-        identifier,
-      },
-    });
-    if (!record) {
-      throw new NotFoundException();
-    }
-    return record;
+    return this.findByIdentifierOrThrowWithClient(identifier, this.prisma);
   }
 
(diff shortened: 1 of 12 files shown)


src/assets/assets.service.ts
```diff
@@ -20,7 +23,19 @@ export class AssetsService {
...
+        identifier,
+      },
+    });
+    if (!record) {
+      throw new NotFoundException();
+    }
+    return record;
+  }
+
+  async findByIdentifierOrThrowWithClient(
```
jowparks, line 38: Can we merge these two methods and use a default instead of having two?

jowparks (approved):
Lgtm one minor comment

src/assets/assets.service.ts
rohanjadvani: I'd like to keep as is since we follow this pattern across other services in our API and the method name is explicit to indicate when a transaction client might be needed. I'm open to changing it but would probably want to have that discussion as a separate change - what do you think?

jowparks: Fine with me, I thought I remember seeing the opposite pattern in the code elsewhere, but if this is the standard then yea lets keep it this way

PR #1233, opened 2023-02-15
feat(api): Bump transaction and change asset increment for debugging

rohanjadvani:
## Summary

Temporarily bump up transaction timeout to debug one block being stuck

## Testing Plan

N/A

## Breaking Change

```
[ ] Yes
```


dgca (approved):

src/blocks-transactions-loader/blocks-transactions-loader.ts
```diff
@@ -127,7 +127,7 @@ export class BlocksTransactionsLoader {
         {
           // We increased this from the default of 5000 because the transactions were
           // timing out and failing to upsert blocks
-          timeout: 30000,
+          timeout: 40000,
```
jowparks, line 130: Want to add an env var here so that we can scale back down without PR?

rohanjadvani: Yeah good idea

src/assets/assets.service.ts
```diff
@@ -63,7 +63,9 @@ export class AssetsService {
   ): Promise<Asset> {
     return prisma.asset.update({
       data: {
```
jowparks, line 65: This change is intentional? Seems like it is accomplishing the same thing as before

rohanjadvani: Yeah it is intentional. I think there's a weird bug in the Prisma engine, so I'm trying this operation instead of addition on the TS side so the DB runner can deal with it

PR #1234, opened 2023-02-15
feat(api): Add asset descriptions controller

rohanjadvani:
## Summary

Add route to list asset descriptions

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/asset-descriptions/asset-descriptions.controller.spec.ts
+++ b/src/asset-descriptions/asset-descriptions.controller.spec.ts
@@ -1,0 +1,137 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { HttpStatus, INestApplication } from '@nestjs/common';
+import { AssetDescriptionType } from '@prisma/client';
+import request from 'supertest';
+import { v4 as uuid } from 'uuid';
+import { AssetsService } from '../assets/assets.service';
+import { PrismaService } from '../prisma/prisma.service';
+import { bootstrapTestApp } from '../test/test-app';
+import { TransactionsService } from '../transactions/transactions.service';
+import { AssetDescriptionsService } from './asset-descriptions.service';
+
+describe('AssetDescriptionsController', () => {
+  let app: INestApplication;
+  let assetDescriptionsService: AssetDescriptionsService;
+  let assetsService: AssetsService;
+  let prisma: PrismaService;
+  let transactionsService: TransactionsService;
+
+  beforeAll(async () => {
+    app = await bootstrapTestApp();
+    assetDescriptionsService = app.get(AssetDescriptionsService);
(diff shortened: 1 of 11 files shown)


src/asset-descriptions/asset-descriptions.controller.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+            {
+              fee: 0,
+              hash: uuid(),
+              notes: [],
+              size: 0,
+              spends: [],
+            },
+          ])
+        )[0];
+
```
jowparks, line 67: Can you add one more transaction/asset/assetdescription and then set a pagination limit to 1 so that we can see `hasNext` is functioning properly?

src/asset-descriptions/asset-descriptions.controller.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const serializedData: SerializedAssetDescription[] = [];
+    for (const record of data) {
+      const transaction = await this.transactionsService.findOrThrow(
+        record.transaction_id,
+      );
+      serializedData.push({
+        object: 'asset_description',
+        id: record.id,
+        transaction_hash: transaction.hash,
```
jowparks, line 58: I take it this is snake cased because `PaginatedList` has some snake casing 😿

src/asset-descriptions/asset-descriptions.controller.spec.ts
rohanjadvani: Added

src/asset-descriptions/asset-descriptions.controller.ts
rohanjadvani: Yeah. We use snake case for our controller responses. There was an API style guide that was used to base this decision when first writing this code, so for consistency using that

src/asset-descriptions/asset-descriptions.service.ts
```diff
@@ -50,4 +57,65 @@ export class AssetDescriptionsService {
...
+      asset_id: options.assetId,
+    };
+    const data = await this.prisma.assetDescription.findMany({
+      cursor,
+      orderBy,
+      skip,
+      take: limit,
+      where,
+    });
+
```
jowparks, line 83: Can we dump this method below and just use a query to get the metadata like

`select min(id), max(id) where ${where} order by ${orderBy}`

Then you can just look to make sure the min < firstItem and max is > lastItem.id


Took me a bit to understand what was happening below with the method, it is doing something pretty simple but with some pretty sophisticated tools (`findMany` and `AssetDescription[]`)

jowparks: Up to you if you want to do this though

jowparks (approved):

src/asset-descriptions/asset-descriptions.service.ts
rohanjadvani: Yeah, let me look into this for a subsequent PR. We're doing this for pagination across the API, but I'll see if there's a simpler pattern

PR #1235, opened 2023-02-15
feat(api): Move asset loading to Graphile

rohanjadvani:
## Breaking Change

```
[ ] Yes
```


src/assets-loader/assets-loader.spec.ts
```diff
@@ -90,55 +74,60 @@ describe('AssetsLoader', () => {
...
-            {
-              identifier: mint.id,
-              metadata: mint.metadata,
-              name: mint.name,
-              owner: mint.owner,
-            },
-            transaction,
-            prisma,
-          );
+          expect(assetsUpsert.mock.calls[0][0]).toEqual({
```
jowparks, line 79: These mock calls look crazy, is it because of the graphile worker? Not a blocker for shipping if the tests pass but we might want to relabel afterwords

src/assets-loader/assets-loader.jobs.controller.ts
```diff
@@ -0,0 +1,25 @@
...
+export class AssetsLoaderJobsController {
+  constructor(private readonly assetsLoader: AssetsLoader) {}
+
+  @MessagePattern(GraphileWorkerPattern.LOAD_ASSET_DESCRIPTIONS)
+  @UseFilters(new GraphileWorkerException())
+  async loadDescriptions({
+    main,
+    transaction,
+  }: LoadDescriptionsOptions): Promise<GraphileWorkerHandlerResponse> {
+    await this.assetsLoader.loadDescriptions(main, transaction);
```
jowparks, line 22: What happens if one of these jobs fails? Does graphile autorequeue a job if it fails because of an exception, overflow etc? I forget

src/blocks-transactions-loader/blocks-transactions-loader.ts
```diff
@@ -132,6 +132,16 @@ export class BlocksTransactionsLoader {
           timeout: this.config.get<number>('BLOCK_LOADER_TRANSACTION_TIMEOUT'),
         },
       );
+
+      for (const transaction of block.transactions) {
+        await this.graphileWorkerService.addJob<LoadDescriptionsOptions>(
```
jowparks, line 137: I wonder if this would be better to queue up on a per block basis. If we have 80 transactions per block we will get 80 jobs per minute, most of that time will be spent handling spin up/down of graphile workers, no? I am not sure it will be an issue and I won't block on this, but just want to be cautious

jowparks (approved):
Seems fine, a little concerned about the transactions vs block level jobs, but I trust your judgement with that

src/assets-loader/assets-loader.jobs.controller.ts
rohanjadvani: We retry with exponential backoff

src/assets-loader/assets-loader.spec.ts
rohanjadvani: It's due to Jest OOM object comparison. Going to leave as is for now

src/blocks-transactions-loader/blocks-transactions-loader.ts
rohanjadvani: I don't think there will be that much overhead since the worker is constantly running and just polling for jobs via the queue. I will monitor and adjust if this gets too slow though

jowparks: Yea we ran into something similar with the multiasset phase 3 changes, trying to recall exactly what it was. The issue may have been fixed by the `graphile` upgrade Jason did here:
https://github.com/iron-fish/ironfish-api/pull/1098

PR #1246, opened 2023-02-21
feat(api): Return timestamps for asset descriptions

rohanjadvani:
## Summary

Return associated block timestamps for asset transactions

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/asset-descriptions/interfaces/serialized-asset-description.ts
+++ b/src/asset-descriptions/interfaces/serialized-asset-description.ts
@@ -9,5 +9,5 @@ export interface SerializedAssetDescription {
   transaction_hash: string;
   type: AssetDescriptionType;
   value: string;
-  timestamp?: string;
+  block_timestamp?: string;
 }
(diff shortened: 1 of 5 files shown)


src/asset-descriptions/interfaces/serialized-asset-description.ts
```diff
@@ -6,7 +6,12 @@ import { AssetDescriptionType } from '@prisma/client';
 export interface SerializedAssetDescription {
   object: 'asset_description';
   id: number;
+  transaction_hash: string;
   type: AssetDescriptionType;
   value: string;
-  transaction_hash: string;
+}
+
```
jowparks, line 13: Instead of extending this type, can we just make timestamp optional?

src/asset-descriptions/utils/asset-descriptions.translator.ts
```diff
@@ -16,3 +19,18 @@ export function serializedAssetDescriptionFromRecord(
...
+
+export function serializedAssetDescriptionWithTimestampFromRecord(
+  assetDescription: AssetDescription,
+  transaction: Transaction,
+  block: Block,
+): SerializedAssetDescriptionWithTimestamp {
+  return {
+    object: 'asset_description',
+    id: assetDescription.id,
+    timestamp: block.timestamp.toISOString(),
```
jowparks, line 31: timestamp is a bit ambigous here, can we name this `block_timestamp`?

jowparks (approved):
Two minor nits, take it or leave it 😸

PR #1248, opened 2023-02-21
feat(api): Return created timestamp for asset

rohanjadvani:
## Summary

Return created block timestamp for assets

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/assets/assets.controller.spec.ts
+++ b/src/assets/assets.controller.spec.ts
@@ -55,6 +55,38 @@ describe('AssetsController', () => {
       });
     });
 
+    describe('when the asset has not been added to a block', () => {
+      it('returns a 404', async () => {
+        const transaction = (
+          await transactionsService.createMany([
+            {
+              fee: 0,
+              hash: uuid(),
+              notes: [],
+              size: 0,
+              spends: [],
+            },
+          ])
+        )[0];
+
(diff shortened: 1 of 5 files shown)


src/assets/assets.controller.ts
```diff
@@ -69,9 +82,20 @@ export class AssetsController {
       const transaction = await this.transactionsService.findOrThrow(
         asset.created_transaction_id,
       );
+
```
jowparks, line 85: Lame that we have to search through blocks just to get the timestamp, but also doesn't seem worth it to write a backfill/update to transaction model for timestamp.

rohanjadvani: Yeah, I'll do that in a subsequent change without us needing to do a reset

src/assets/assets.controller.spec.ts
```diff
@@ -125,6 +158,12 @@ describe('AssetsController', () => {
           },
         ])
       )[0];
```
jowparks, line 160: Can you add a test for the `find` where you get a failure because block isn't found? Also I think we could add a an error message to that exception saying block wasn't found for where asset was created, might be helpful if we ever have to debug.

jowparks (approved):

PR #1249, opened 2023-02-22
Add an index to faucet and remove transaction

NullSoldier:
## Summary

This API is an optional debugging API. I'm moving it to the read only client, removing the transaction (since its on a follower anyway).

Data dog is showing this as the second slowest API, because the table is growing very large.

```
count: 339645
completed: 339647
```

Slowest DB Query
```
SELECT "public"."faucet_transactions"."id",
         "public"."faucet_transactions"."created_at",
         "public"."faucet_transactions"."updated_at",
         "public"."faucet_transactions"."email",
         "public"."faucet_transactions"."public_key",
         "public"."faucet_transactions"."started_at",
         "public"."faucet_transactions"."completed_at",
         "public"."faucet_transactions"."tries",
         "public"."faucet_transactions"."hash"
FROM "public"."faucet_transactions"
WHERE ("public"."faucet_transactions"."started_at" IS NOT NULL
        AND "public"."faucet_transactions"."completed_at" IS NULL)
ORDER BY  "public"."faucet_transactions"."created_at" ASC LIMIT $1 OFFSET $2
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/prisma/migrations/20230222171607_add_faucet_index/migration.sql
+++ b/prisma/migrations/20230222171607_add_faucet_index/migration.sql
@@ -1,11 +1,5 @@
--- DropIndex
-DROP INDEX "index_users_on_graffiti";
-
 -- CreateIndex
 CREATE INDEX "index_faucet_transactions_on_completed_at" ON "faucet_transactions"("completed_at");
 
 -- CreateIndex
 CREATE INDEX "index_faucet_transactions_on_completed_at_and_started_at" ON "faucet_transactions"("started_at", "completed_at");
-
--- CreateIndex
-CREATE INDEX "index_users_on_graffiti" ON "users"("graffiti");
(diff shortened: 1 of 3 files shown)


prisma/migrations/20230222171607_add_faucet_index/migration.sql
```diff
@@ -0,0 +1,11 @@
+-- DropIndex
+DROP INDEX "index_users_on_graffiti";
```
rohanjadvani, line 2: Can you remove the changes for graffiti on `users`? We have a trigram index on this, but Prisma is regenerating it

NullSoldier: done

rohanjadvani (approved):

PR #1252, opened 2023-02-24
feat(api): Return asset identifier and name with descriptions

rohanjadvani:
## Summary

Return asset info along with descriptions

## Testing Plan

Manual testing

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

jowparks (commented):
Looks fine, maybe add a test or two?

NullSoldier:
Joe is right we should add a test but this is blocking dan from shipping the new block explorer so we need to get this out.

PR #1257, opened 2023-02-26
feat: add redemption model and service for kyc (1/3)

jowparks:
## Summary
Adds 
1. redemption model
2. redemption service
3. redemption controller
  a. `POST redemption` with `public_address` to create new redemption for user
  b. `GET redemption` to retrieve serialized redemption for user

for upcoming KYC/airdrop to users for testnet participation.
## Testing Plan
Tests pass
## Breaking Change

```
[ ] Yes
```


src/events/multi-asset.upsert.service.spec.ts
```diff
@@ -259,7 +259,7 @@ describe('MultiAssetUpsertService', () => {
         block: {
           hash: 'multiassetupsertblockhash1',
           previousBlockHash: head.block_hash,
```
jowparks, line 261: Failing test due to end of testnet passed

src/telemetry/telemetry.controller.spec.ts
```diff
@@ -66,6 +66,7 @@ describe('TelemetryController', () => {
     usersService = app.get(UsersService);
     versionsService = app.get(VersionsService);
     await app.init();
```
jowparks, line 68: These test are failing as a result of this conditional check:
https://github.com/iron-fish/ironfish-api/pull/1256/files#diff-b9ee66213b9bfe61d1f85d4cba47efe234bf537c6c9b05f1b072dfbb1d9a4654R113
Mocking test time to make tests pass

src/redemptions/redemption.controller.ts
```diff
@@ -0,0 +1,79 @@
...
+  constructor(private readonly redemptionService: RedemptionService) {}
+
+  matchUserOrThrow(id: number, user: User): void {
+    if (id !== user.id) {
+      throw new ForbiddenException('User id does not match magic link user');
+    }
+  }
+
+  @ApiExcludeEndpoint()
+  @Post(':id')
```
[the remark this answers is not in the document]
jowparks: Good call, I saw this pattern in a different API we have, but what you are saying makes total sense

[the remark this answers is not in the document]
jowparks: https://github.com/iron-fish/ironfish-api/blob/b4e4acd52b8a24bffb586062d0d77a706d2c4def/src/users/users.controller.ts#L347-L349

src/redemptions/redemption.controller.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+      email: faker.internet.email(),
+      graffiti: uuid(),
+      countryCode: faker.address.countryCode('alpha-3'),
+    });
+    jest
+      .spyOn(magicLinkService, 'getEmailFromHeader')
+      .mockImplementation(() => Promise.resolve(user.email));
+    return user;
+  };
+
```
jowparks, line 45: example Redemption creation

src/redemptions/redemption.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+    @Context() { user }: MagicLinkContext,
+    @Body(
+      new ValidationPipe({
+        errorHttpStatusCode: HttpStatus.UNPROCESSABLE_ENTITY,
+      }),
+    )
+    dto: RedemptionDto,
+  ): Promise<SerializedRedemption> {
+    if (await this.redemptionService.find(user)) {
+      throw new ForbiddenException('Redemption already exists');
```
rohanjadvani, line 43: 403 should be used if a user is authenticated but they don't have permissions for this operation. We're using 422 for user creation [here](https://github.com/iron-fish/ironfish-api/blob/master/src/users/users.service.ts#L131-L138). We should keep this consistent

src/redemptions/redemption.controller.ts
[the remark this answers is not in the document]
rohanjadvani: I think this pattern is standard for REST-ful updates. I think most updates are done via `PUT <resource>/:id`. For creation and retrieval, we can skip passing an ID

src/redemptions/redemption.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+  async create(
+    @Context() { user }: MagicLinkContext,
+    @Body(
+      new ValidationPipe({
+        errorHttpStatusCode: HttpStatus.UNPROCESSABLE_ENTITY,
+      }),
+    )
+    dto: RedemptionDto,
+  ): Promise<SerializedRedemption> {
+    if (await this.redemptionService.find(user)) {
```
rohanjadvani, line 42: I think it'd save us some Jumio calls if we check points before creating redemptions for a user. What do you think?

rohanjadvani: This will probably require aggregating points across phases.

prisma/schema.prisma
```diff
@@ -51,15 +51,34 @@ model User {
...
+  created_at       DateTime     @default(now()) @db.Timestamp(6)
+  updated_at       DateTime     @default(now()) @updatedAt @db.Timestamp(6)
+  user             User         @relation(fields: [user_id], references: [id])
+  user_id          Int          @unique
+  kyc_attempts     Int          @default(0)
+  kyc_status       KycStatus
+  jumio_account_id String?      @db.VarChar
+  public_address   String       @db.VarChar
+  email_verified   Boolean      @default(false)
+  dedupe_status    DedupeStatus
```
rohanjadvani, line 76: I don't think this should live on the kyc rows. We can add eligibility to the user directly

prisma/migrations/20230226191138_add_redemption_kyc/migration.sql
```diff
@@ -0,0 +1,30 @@
+-- CreateEnum
+CREATE TYPE "KycStatus" AS ENUM ('NOT_STARTED', 'PASS', 'PENDING', 'FAIL_TRANSACTION', 'FAIL_MAX_ATTEMPTS');
+
+-- CreateEnum
+CREATE TYPE "DedupeStatus" AS ENUM ('NOT_STARTED', 'PASS', 'FAIL');
+
+-- CreateTable
+CREATE TABLE "redemptions" (
```
NullSoldier, line 8: we could just call this `kyc`

prisma/migrations/20230226191138_add_redemption_kyc/migration.sql
```diff
@@ -0,0 +1,30 @@
...
+    "id" SERIAL NOT NULL,
+    "created_at" TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated_at" TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "user_id" INTEGER NOT NULL,
+    "kyc_attempts" INTEGER NOT NULL DEFAULT 0,
+    "kyc_status" "KycStatus" NOT NULL,
```
NullSoldier, line 14: if we call this kyc these 2 should just be `status` and `attempts`

prisma/migrations/20230226191138_add_redemption_kyc/migration.sql
```diff
@@ -0,0 +1,30 @@
...
+    "jumio_account_id" VARCHAR,
+    "public_address" VARCHAR NOT NULL,
+    "email_verified" BOOLEAN NOT NULL DEFAULT false,
+    "dedupe_status" "DedupeStatus" NOT NULL,
```
NullSoldier, line 18: delete this, it's unrelated to this model

prisma/migrations/20230226191138_add_redemption_kyc/migration.sql
```diff
@@ -0,0 +1,30 @@
+-- CreateEnum
+CREATE TYPE "KycStatus" AS ENUM ('NOT_STARTED', 'PASS', 'PENDING', 'FAIL_TRANSACTION', 'FAIL_MAX_ATTEMPTS');
+
+-- CreateEnum
```
NullSoldier, line 4: And remove this

prisma/migrations/20230226191138_add_redemption_kyc/migration.sql
rohanjadvani: I think we should have this live on the user

prisma/migrations/20230226191138_add_redemption_kyc/migration.sql
```diff
@@ -0,0 +1,30 @@
...
+CREATE TYPE "KycStatus" AS ENUM ('NOT_STARTED', 'PASS', 'PENDING', 'FAIL_TRANSACTION', 'FAIL_MAX_ATTEMPTS');
```
NullSoldier, line 2: Nothing should ever be NOT_STARTED right?

prisma/schema.prisma
```diff
@@ -51,15 +51,34 @@ model User {
...
   github              String?     @db.VarChar
   hashed_ip_address   String?     @db.VarChar
-  events              Event[]
-  node_uptime         NodeUptime?
-  points              UserPoints?
+  redemption          Redemption?
+
+  events      Event[]
+  node_uptime NodeUptime?
+  points      UserPoints?
```
NullSoldier, line 58: revert these

prisma/schema.prisma
```diff
@@ -51,15 +51,34 @@ model User {
...
 
   @@index([email], name: "index_users_on_email")
   @@index([graffiti], name: "index_users_on_graffiti")
   @@map("users")
 }
 
+model Redemption {
```
NullSoldier, line 65: we could just call this kyc

prisma/schema.prisma
```diff
@@ -51,15 +51,34 @@ model User {
...
+  id               Int          @id @default(autoincrement())
+  created_at       DateTime     @default(now()) @db.Timestamp(6)
+  updated_at       DateTime     @default(now()) @updatedAt @db.Timestamp(6)
+  user             User         @relation(fields: [user_id], references: [id])
+  user_id          Int          @unique
+  kyc_attempts     Int          @default(0)
+  kyc_status       KycStatus
```
NullSoldier, line 72: if we call this kyc these 2 should just be status and attempts

prisma/schema.prisma
```diff
@@ -51,15 +51,34 @@ model User {
...
+  jumio_account_id String?      @db.VarChar
+  public_address   String       @db.VarChar
+  email_verified   Boolean      @default(false)
```
NullSoldier, line 75: This is already on the user as `last_login_at`

prisma/schema.prisma
NullSoldier: Yes, I think we should delete this its unrelated to this model.

src/redemptions/redemption.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+import { Context } from '../common/decorators/context';
+import { MagicLinkContext } from '../common/interfaces/magic-link-context';
+import { RedemptionDto } from './dto/redemption.dto';
+import { SerializedRedemption } from './interfaces/serializedRedemption';
+import { RedemptionService } from './redemption.service';
+import { serializeRedemption } from './utils/serialize-redemption';
+
+@ApiTags('Redemption')
+@Controller('redemption')
+export class RedemptionController {
```
rohanjadvani, line 27: nit for consistency

```suggestion
export class RedemptionsController {
```

prisma/schema.prisma
```diff
@@ -329,3 +349,17 @@ enum AssetDescriptionType {
...
+
+enum KycStatus {
+  NOT_STARTED
+  PASS
+  PENDING
+  FAIL_TRANSACTION
+  FAIL_MAX_ATTEMPTS
+}
+
+enum DedupeStatus {
```
NullSoldier, line 361: We should remove this if we remove it above

prisma/schema.prisma
```diff
@@ -329,3 +349,17 @@ enum AssetDescriptionType {
 
   @@map("asset_description_type")
 }
+
+enum KycStatus {
+  NOT_STARTED
```
NullSoldier, line 354: No redemption should ever be `NOT_STARTED` right?

src/redemptions/dto/redemption.dto.ts
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { IsNotEmpty, IsString } from 'class-validator';
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
```
NullSoldier, line 8: two licenses

src/redemptions/interfaces/serializedRedemption.ts
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { KycStatus } from '@prisma/client';
+
+export interface SerializedRedemption {
+  object: 'redemption';
+  id: number;
+  created_at: string;
```
rohanjadvani, line 9: I don't think we need to return `created` / `updated`

src/redemptions/dto/redemption.dto.ts
```diff
@@ -0,0 +1,13 @@
...
+export class RedemptionDto {
```
NullSoldier, line 9: In other dto's this is named after the action and then the noun, so it should be `CreateRedemptionDto`

src/redemptions/interfaces/serializedRedemption.ts
```diff
@@ -0,0 +1,17 @@
...
+  updated_at: string;
+  user_id: number;
```
rohanjadvani, line 11: We may not need this if the user is already auth'd

src/redemptions/redemption.controller.ts
rohanjadvani: As discussed offline, we can replace with an eligibility flag on `users`

NullSoldier: We don't have the data for points yet, but we can add

* Are they from Iran, or North Korea

src/redemptions/redemption.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @ApiExcludeEndpoint()
+  @Get()
+  @UseGuards(MagicLinkGuard)
+  async retrieve(
+    @Context() { user }: MagicLinkContext,
+  ): Promise<SerializedRedemption | null> {
+    const redemption = await this.redemptionService.find(user);
+    if (!redemption) {
```
NullSoldier, line 59: This is wrong, look at the return type and then look at this error.

src/redemptions/redemption.service.ts
```diff
@@ -0,0 +1,88 @@
...
+      data: {
+        user: { connect: { id: user.id } },
+        public_address,
+        dedupe_status: DedupeStatus.NOT_STARTED,
+        kyc_status: KycStatus.NOT_STARTED,
+      },
+    });
+  }
+
+  async attemptKyc(
```
NullSoldier, line 62: Why doesn't it just create the redemption in the thing that calls this?

prisma/schema.prisma
lwisne: No, they are not the same thing. See spec.

prisma/migrations/20230226191138_add_redemption_kyc/migration.sql
lwisne: Keep in mind a minority of users are going through this process. This avoids having to worry about filling in this field intelligently for the 100k+ users that we're essentially ignoring because they have no points or haven't verified email.

rohanjadvani: There are additional heuristics I had in mind, and it felt weird to check that during redemption vs before a user requests anything (since I thought Mat was going to work on some manual checking). Open to discussing this further, though

NullSoldier: I think we should always check deduperun_status because it duplicates this. If we don't propagate this it when we may make a mistake and they could be out of sync.

NullSoldier: We decided to remove this since it already lives on eligibility_checks.status and we can just join to get it

prisma/schema.prisma
NullSoldier: Spec was updated to remove this

prisma/schema.prisma
NullSoldier: We decided to remove this

prisma/schema.prisma
NullSoldier: Removed

src/redemptions/dto/redemption.dto.ts
NullSoldier: fixed

src/redemptions/dto/redemption.dto.ts
NullSoldier: fixed

src/redemptions/redemption.controller.ts
NullSoldier: fixed

src/redemptions/redemption.controller.ts
NullSoldier: fixed

src/redemptions/redemption.controller.ts
NullSoldier: We'll do this in one of the other PRs

src/telemetry/telemetry.controller.spec.ts
NullSoldier: This was fixed in another PR. I rebased.

PR #1259, opened 2023-02-27
Re-correct phase3 end time

hairtail:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -57,5 +57,5 @@ export const POOL_4_CATEGORIES: Array<EventType> = [
   EventType.MULTI_ASSET_TRANSFER,
 ];
 
-// 2022 Feb 26 12 AM UTC
-export const PHASE_3_END = new Date(Date.UTC(2023, 1, 26, 0, 0, 0));
+// 2023 Feb 26 12 AM UTC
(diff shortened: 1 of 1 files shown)


src/common/constants.ts
```diff
@@ -57,5 +57,5 @@ export const POOL_4_CATEGORIES: Array<EventType> = [
   EventType.MULTI_ASSET_TRANSFER,
 ];
 
-// 2022 Feb 26 12 AM UTC
-export const PHASE_3_END = new Date(Date.UTC(2023, 1, 26, 0, 0, 0));
+// 2023 Feb 26 12 AM UTC
+export const PHASE_3_END = new Date(Date.UTC(2023, 2, 26, 0, 0, 0));
```
NullSoldier, line 61:
```suggestion
export const PHASE_3_END = new Date(Date.UTC(2023, 1, 26, 0, 0, 0));
```

NullSoldier:
```suggestion
export const PHASE_3_END = new Date(Date.UTC(2023, 1, 26, 0, 0, 0));
```

NullSoldier:
The PR is wrong but I'm gonna merge in the comment change.

jowparks:
@hairtail `Date`s are indexed by 0 in JS, confusing, but the value was already correct.

PR #1269, opened 2023-02-28
chore(api): Add flake.nix

rohanjadvani:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,24 @@
+{
+  description = "A prisma test project";
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/master";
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -0,0 +1,24 @@
+{
+  description = "A prisma test project";
```
dguenther, line 2: This could be something more specific to Iron Fish, since I think this will be an Iron Fish flake

dguenther (approved):

flake.nix
rohanjadvani: I forgot to change that last night 😔 I'll fix it

PR #1273, opened 2023-03-01
feat(api): Add `GET /kyc/config`

rohanjadvani:
## Summary

Add config on testnet pools and kyc / airdrop dates

## Testing Plan

Unit test

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-kyc/kyc.controller.ts
+++ b/src/jumio-kyc/kyc.controller.ts
@@ -20,6 +20,7 @@ import { JumioTransactionService } from '../jumio-transactions/jumio-transaction
 import { RedemptionService } from '../redemptions/redemption.service';
 import { CreateKycDto } from './dto/create-kyc.dto';
 import { SerializedKyc } from './interfaces/serialized-kyc';
+import { SerializedKycConfig } from './interfaces/serialized-kyc-config';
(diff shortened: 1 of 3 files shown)


jowparks (approved):
I would just ask you get skylar or Dan to approve, I want verification of the pools/details.

src/jumio-kyc/kyc.controller.ts
```diff
@@ -80,4 +81,41 @@ export class KycController {
...
+          pool_name: 'Phase 2 Pool',
+        },
+        {
+          airdrop_completed_by: new Date(Date.UTC(2023, 2, 26, 0, 0, 0)),
+          coins: 210000,
+          kyc_completed_by: new Date(Date.UTC(2023, 3, 6, 0, 0, 0)),
+          name: 'Phase 3',
+          pool_name: 'Phase 3 Pool',
+        },
+      ],
```
NullSoldier, line 118: Should we key this by pool? I think it'll be easier for dan since he needs to reference them explicitly in his design rather than loop over them.

 - pool1: {}
 - pool2: {}
 - pool3: {}
 - pool4: {}

NullSoldier (approved):

PR #1277, opened 2023-03-02
fix template and make jumio stuff optional

lwisne:
## Breaking Change

```
[ ] Yes
```


.env.template
```diff
@@ -26,7 +26,7 @@ JUMIO_URL=amer-1.jumio.ai/api/v1
 JUMIO_WORKFLOW_DEFINITION=10013
 MAGIC_SECRET_KEY=test
 MAX_DEPOSIT_SIZE=0.6
-MAX_KYC_ATTEMPTS=3
+KYC_MAX_ATTEMPTS=3
```
rohanjadvani, line 29: This change makes sense to me since the validation expects this key

src/app.module.ts
```diff
@@ -95,8 +95,8 @@ export const REST_MODULES = [
         IRONFISH_API_KEY: joi.string().required(),
         IRONFISH_PHASE_ONE_API_URL: joi.string().required(),
         IRONFISH_PHASE_TWO_API_URL: joi.string().required(),
-        JUMIO_API_SECRET: joi.string().required(),
-        JUMIO_API_TOKEN: joi.string().required(),
+        JUMIO_API_SECRET: joi.string().optional().allow(''),
```
rohanjadvani, line 98: I think it'd be weird to spin up the API without the secret / token since we'll need those values to function. We use `test` (ex. `MAGIC_SECRET_KEY=test` in `.env.template`) and then that value can be updated in `.env` if needed. I'd vote to keep this as `required()` but use `test` or something in the template. What do you think?

.env.template
lwisne: yeah, that's why i did it. you get a rejection if you use the template copy

src/app.module.ts
lwisne: ok, that's fine by me. Basically, this is where you would use a staging/dummy DB value, but I don't think we have that. The template value doesn't work so wanted to fixt aht.

rohanjadvani (approved):

NullSoldier (approved):

PR #1278, opened 2023-03-02
feat(api): Return pool points

rohanjadvani:
## Summary

Return cached values from previous pools in metrics request.

I chose to add a new field to the response instead of updating the existing `pools` field because the testnet UI is using that and it felt weird to set the count to NULL since that value isn't populated. It seemed cleaner to just return just the points altogether, but open to changing.

## Testing Plan

Unit test

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/users/interfaces/serialized-user-metrics.ts
+++ b/src/users/interfaces/serialized-user-metrics.ts
@@ -28,4 +28,7 @@ export interface SerializedUserMetrics {
     total_hours: number;
     last_checked_in: string | null;
   };
+  pool_points?: {
+    [pool: string]: number;
+  };
 }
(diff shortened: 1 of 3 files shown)


src/users/interfaces/serialized-user-metrics.ts
```diff
@@ -28,4 +28,7 @@ export interface SerializedUserMetrics {
     total_hours: number;
     last_checked_in: string | null;
   };
+  pool_points?: {
+    [pool: string]: number;
```
jowparks, line 32: Do we want to just explicitly type out the pools here, will be easier to understand if someone has to come back to this in the future. We will never add new pools, right?

jowparks (approved):

src/users/interfaces/serialized-user-metrics.ts
rohanjadvani: Yeah sounds good

PR #1279, opened 2023-03-02
Change check to use pool points

NullSoldier:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/redemptions/redemption.service.ts
+++ b/src/redemptions/redemption.service.ts
@@ -174,7 +174,13 @@ export class RedemptionService {
   ): Promise<string | null> {
     const points = await this.userPointsService.findOrThrow(user.id, prisma);
 
-    if (points.total_points === 0) {
+    const hasPoints =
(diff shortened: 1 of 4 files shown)


src/redemptions/redemption.service.ts
```diff
@@ -174,7 +174,13 @@ export class RedemptionService {
...
     const points = await this.userPointsService.findOrThrow(user.id, prisma);
 
-    if (points.total_points === 0) {
+    const hasPoints =
+      !points.pool1_points &&
+      !points.pool2_points &&
+      !points.pool3_points &&
+      !points.pool4_points;
+
+    if (hasPoints) {
```
lwisne, line 183: this is the opposite of what it would suggest. I would call this `hasNoPoints`

NullSoldier: Fixed

jowparks (approved):

PR #1281, opened 2023-03-03
feat: stores id info and failure reason in database

jowparks:
## Summary
Stores country info on redemption and determines if user is banned based on id
## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

src/redemptions/redemption.service.ts
```diff
@@ -3,12 +3,19 @@
...
 import { DecisionStatus, KycStatus, Redemption, User } from '@prisma/client';
+import { instanceToPlain } from 'class-transformer';
 import { ApiConfigService } from '../api-config/api-config.service';
 import { JumioTransactionRetrieveResponse } from '../jumio-api/interfaces/jumio-transaction-retrieve';
+import { IdDetails } from '../jumio-kyc/kyc.service';
 import { PrismaService } from '../prisma/prisma.service';
 import { BasePrismaClient } from '../prisma/types/base-prisma-client';
 import { UserPointsService } from '../user-points/user-points.service';
 
+export type CalculatedStatus = {
```
NullSoldier, line 14: this is only used in one place, can you inline it unless we use it in more than one place?

PR #1286, opened 2023-03-03
Send back eligibility in can_attempt

NullSoldier:
## Summary

This will not send back if you can attempt a KYC flow, now it will send back if you are eligible.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-kyc/kyc.controller.ts
+++ b/src/jumio-kyc/kyc.controller.ts
@@ -63,17 +63,16 @@ export class KycController {
   ): Promise<SerializedKyc | { can_attempt: boolean }> {
     const redemption = await this.redemptionService.find(user);
 
-    const canAttemptError = await this.redemptionService.isEligible(user);
-    const canAttempt = !canAttemptError;
+    const eligibleError = await this.redemptionService.isEligible(user);
 
     if (!redemption) {
-      return { can_attempt: canAttempt };
+      return { can_attempt: !eligibleError };
     }
 
     const jumioTransaction =
       await this.jumioTransactionService.findLatestOrThrow(user);
 
-    return serializeKyc(redemption, jumioTransaction, canAttempt);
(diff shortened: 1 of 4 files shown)


src/redemptions/redemption.service.ts
```diff
@@ -167,8 +167,7 @@ export class RedemptionService {
     });
   }
 
-  async canAttempt(
-    redemption: Redemption | null,
+  async isEligible(
```
rohanjadvani, line 170: nit: it's a bit odd that the method name is `isEligible` but is returning an error or null. It might be easier to follow if we return a boolean along with an error reason, similar to how we do in the node.

src/redemptions/redemption.service.ts
```diff
@@ -192,6 +191,20 @@ export class RedemptionService {
...
 
+    return null;
+  }
+
+  async canAttempt(
+    redemption: Redemption | null,
+    user: User,
+    prisma?: BasePrismaClient,
+  ): Promise<string | null> {
+    const eligibleError = await this.isEligible(user, prisma);
```
rohanjadvani, line 202: nit: same comment as above. Reads a little weird `eligibleError = isEligible(...)`

src/jumio-kyc/kyc.controller.ts
```diff
@@ -63,21 +63,16 @@ export class KycController {
...
-      redemption,
-      user,
-    );
-
-    const canAttempt = !canAttemptError;
+    const eligibleError = await this.redemptionService.isEligible(user);
 
     if (!redemption) {
-      return { can_attempt: canAttempt };
+      return { can_attempt: !eligibleError };
```
rohanjadvani, line 69: Same as other comments, might be good to return a message that can be passed to the UI

rohanjadvani (approved):
Approving in case we want to deploy/merge this soon, but there's some clean up I'd like to be addressed at some point

PR #1288, opened 2023-03-03
Store callback JSON in the DB

NullSoldier:
## Summary

This is useful for debugging

 - rename `JumioTransaction.latest_callback_at` -> `JumioTransaction.last_callback_at`
 - Add `JumioTransaction.latest_callback`

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):

lwisne:
Is this callback one with PII? Just want to be very careful with this so we're sure we're not creating a massive DB of government IDs and photos, etc.

NullSoldier:
the callback has no PII in it, the PII is in the fetch for the workflow.

PR #1294, opened 2023-03-04
Move banned countries into config

NullSoldier:
## Summary

This lets us tune it and test on staging, and it helps to hide what countries are actually banned.

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):
LGTM, verified it's in Heroku

PR #1297, opened 2023-03-04
feat(api): Add refresh queues for user points

rohanjadvani:
## Summary

Throttle some requests when backfilling points due to connection limits on our old dynos

## Testing Plan

N/A

## Breaking Change

```
[ ] Yes
```


jowparks (commented):
Why do we need this? There shouldn't be any new points coming in right?

Also won't this make it take up more workers rather than throttling? Wouldn't you want to just limit the number of workers for your queue instead?

PR #1298, opened 2023-03-04
fix: update logic for redemption/attempt checks

jowparks:
## Summary
Small updates to logic in kyc endpoints. Please check logic

## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-kyc/kyc.controller.spec.ts
+++ b/src/jumio-kyc/kyc.controller.spec.ts
@@ -217,7 +217,7 @@ describe('KycController', () => {
   });
 
   describe('GET /kyc', () => {
-    it('retrieves kyc info when it exists', async () => {
+    it('retrieves kyc info when it exists and shows user is not eligible to start new kyc', async () => {
       const user = await mockUser();
       const { redemption, transaction } = await kycService.attempt(user, 'foo');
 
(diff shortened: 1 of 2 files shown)


src/jumio-kyc/kyc.controller.ts
```diff
@@ -47,15 +47,8 @@ export class KycController {
     )
     dto: CreateKycDto,
   ): Promise<SerializedKyc> {
-    const { redemption, transaction } = await this.kycService.attempt(
-      user,
-      dto.public_address,
-    );
-
```
jowparks, line 54: `canAttempt` is called inside of `attempt` method.

src/jumio-kyc/kyc.controller.ts
```diff
@@ -70,16 +63,19 @@ export class KycController {
   > {
     const redemption = await this.redemptionService.find(user);
```
jowparks, line 65: `isEligible` is only a subset of `canAttempt`, updated to include all checks

src/jumio-kyc/kyc.controller.ts
jowparks: nvm I see why you did this

src/jumio-kyc/kyc.controller.spec.ts
```diff
@@ -227,7 +227,12 @@ describe('KycController', () => {
         .expect(HttpStatus.OK);
 
       expect(body).toMatchObject(
```
jowparks, line 229: If the user has a kyc that is in progress, then when the `GET` endpoint is hit `attemptable` should be false, no?

jowparks:
Closing this since it seems this `can_attempt` flag is overloaded for eligibility. Seems to be working in UI, so I am gonna just let it be

PR #1300, opened 2023-03-04
Add jobs to refresh redemptions

NullSoldier:
## Summary

This will allow us to refresh jobs that are stale.

## Breaking Change

```
[ ] Yes
```


src/jumio-kyc/kyc.jobs.controller.ts
```diff
@@ -0,0 +1,69 @@
...
+    private readonly loggerService: LoggerService,
+    private readonly usersService: UsersService,
+    private readonly kycService: KycService,
+    private readonly jumioTransactionService: JumioTransactionService,
+    private readonly redemptionService: RedemptionService,
+  ) {}
+
+  @MessagePattern(GraphileWorkerPattern.REFRESH_USERS_REDEMPTION)
+  @UseFilters(new GraphileWorkerException())
+  async refreshUsersPoints(): Promise<GraphileWorkerHandlerResponse> {
```
jowparks, line 31: nit: rename functions

src/jumio-kyc/kyc.jobs.controller.ts
```diff
@@ -0,0 +1,69 @@
...
+      await this.kycService.refreshUser(user);
+    }
+
+    return { requeue: false };
+  }
+
+  private async *usersGenerator(): AsyncGenerator<User> {
+    let after: number | undefined;
+    let hasNext = false;
+    do {
```
jowparks, line 56: You could filter users by those with points, right?

jowparks (approved):

src/jumio-kyc/kyc.jobs.controller.ts
NullSoldier: We avoid users down the line that don't have a redemption and that avoids an extra query

PR #1303, opened 2023-03-04
Send back kyc_max_attempts in serializedKYC

NullSoldier:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-kyc/kyc.controller.spec.ts
+++ b/src/jumio-kyc/kyc.controller.spec.ts
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { HttpStatus, INestApplication } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
 import { DecisionStatus, KycStatus, User } from '@prisma/client';
 import assert from 'assert';
 import axios from 'axios';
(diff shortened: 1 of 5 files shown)


src/jumio-kyc/kyc.controller.spec.ts
```diff
@@ -39,6 +41,7 @@ describe('KycController', () => {
   beforeAll(async () => {
     app = await bootstrapTestApp();
     prisma = app.get(PrismaService);
+    config = app.get(ConfigService);
```
rohanjadvani, line 44:
```suggestion
    config = app.get(ApiConfigService);
```

NullSoldier: Fixed

src/jumio-kyc/utils/serialize-kyc.ts
```diff
@@ -12,12 +13,17 @@ export function serializeKyc(
   canAttemptReason: string,
   canCreate: boolean,
   canCreateReason: string,
+  config: ApiConfigService,
 ): SerializedKyc {
   assert.ok(redemption.jumio_account_id);
 
+  const maxAttempts =
```
rohanjadvani, line 20: Feels a little weird to pass an injected dependency into a serializer. I think it'd be cleaner if we did this in the caller (or just passing the config value as a default)

rohanjadvani (approved):

src/jumio-kyc/utils/serialize-kyc.ts
NullSoldier: That is a better way but it it requires every test to change that tests this to do that

PR #1315, opened 2023-03-06
feat: gate airdrop eligibility on users latest eligble pool

jowparks:
## Summary
Users should only be able to complete KYC up until the last eligibility date for their pools
## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```


src/redemptions/redemption.service.ts
```diff
@@ -246,6 +255,34 @@ export class RedemptionService {
     return { eligible: true, reason: '' };
   }
 
+  currentDate = (): Date => {
```
rohanjadvani, line 258: Is the only reason we have this to mock out in the tests? There should be a way to mock the time via Jest or the Date constructor without needing this.

src/redemptions/redemption.service.ts
```diff
@@ -246,6 +255,34 @@ export class RedemptionService {
...
+    return new Date();
+  };
+
+  userDeadline = async (userId: number): Promise<Date> => {
```
rohanjadvani, line 262:
```suggestion
  async userDeadline(userId: number): Promise<Date> {
```

For consistency

src/jumio-kyc/config/airdrop.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+    {
+      airdrop_completed_by: new Date(Date.UTC(2023, 3, 6, 0, 0, 0)),
+      coins: 210000,
+      kyc_completed_by: new Date(Date.UTC(2023, 2, 26, 0, 0, 0)),
+      name: 'pool_four',
+      pool_name: 'Phase 3 Pool',
+    },
+  ],
+};
```
NullSoldier, line 35: there is already constants.ts

src/jumio-kyc/config/airdrop.ts
```diff
@@ -0,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+export const AIRDROP_CONFIG = {
```
rohanjadvani, line 4: Move to constants file

src/redemptions/redemption.service.ts
```diff
@@ -246,6 +255,34 @@ export class RedemptionService {
...
+    }
+    const pool3 = AIRDROP_CONFIG.data.find((c) => c.name === 'pool_three');
+    if (pool3 && userPoints.pool3_points !== 0) {
+      eligiblePools.push(pool3);
+    }
+    const pool4 = AIRDROP_CONFIG.data.find((c) => c.name === 'pool_four');
+    if (pool4 && userPoints.pool4_points !== 0) {
+      eligiblePools.push(pool4);
+    }
+    return new Date(
```
rohanjadvani, line 281: Note this will return `-Inf` if no eligible pools are present for a `userId`

src/redemptions/redemption.service.ts
jowparks: Seems like it could be flakey though if there is a different Date constructor call that is introduced in future. Also could be flakey if we mock all dates and some other check relies on date constructor.

src/redemptions/redemption.service.ts
jowparks: We can definitely handle this explicitly, but user can't  create redemption if they don't have points in one of phases.

rohanjadvani: Yeah, I think it's fine as is. Just was a note while reading this

src/redemptions/redemption.service.ts
rohanjadvani: If something is based on the current live time, then I think it's reasonable to mock the system time via Jest. All Date constructors then should be consistent

jowparks: for reference I tried:
`jest.useFakeTimers().setSystemTime(new Date('2090-03-06'));`

rohanjadvani:
```suggestion
  currentDate(): Date {
```

Discussed offline, but it's behaving a bit weird. I prefer just mocking the `Date` constructor since I think that's cleaner than having a `currentDate` on a service

rohanjadvani (approved):

PR #1321, opened 2023-03-06
fix: default to submitted status rather than try again

jowparks:
## Summary
If we bypass all the checks in `calculateStatus` we currently default to `TRY_AGAIN`. This caused issue in a tested transaction for a user.

In order to make this switch we need to verify the blocking logic for the `TRY_AGAIN`/`FAILURE` statuses here is sufficient.
## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/redemptions/redemption.service.ts
+++ b/src/redemptions/redemption.service.ts
@@ -93,7 +93,7 @@ export class RedemptionService {
     }
 
     return {
-      status: KycStatus.TRY_AGAIN,
+      status: KycStatus.SUBMITTED,
       failureMessage: null,
       idDetails,
     };


src/redemptions/redemption.service.ts
```diff
@@ -93,7 +93,7 @@ export class RedemptionService {
     }
 
     return {
-      status: KycStatus.TRY_AGAIN,
+      status: KycStatus.SUBMITTED,
```
NullSoldier, line 96: This is not a blocking comment but this is a dangerous change if you haven't carefully checked every case of your above checks. It's fine but if we have bugs users will pass KYC when they should fail. That's the main issue here.

I would generally assume all failures and whitelist specific cases rather than black list failure it's much safer.

NullSoldier (approved):

src/redemptions/redemption.service.ts
lwisne: yes, this is the approach we chose. 

The more specific solution here will be https://airtable.com/appIXmGgVqP9QdbCf/tblAWSTOQbTq6vh6X/viwfithULCYLCL6B6/recAsHBMu6gviOgZq?blocks=hide

lwisne (changes requested):
my understanding is we are closing this. LMK if you don't agree.

jowparks:
Correct we are closing this

PR #1322, opened 2023-03-06
fix: banned countries into code rather than env var

jowparks:
## Summary
Currently we can update banned countries via env var, this could be a problem for auditing, tracability, and review process for banning. Moving into code as constant.
## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```


lwisne (approved):

NullSoldier (approved):

rohanjadvani:
Can you remove this from the `joi` validation? Looks good otherwise.

PR #1323, opened 2023-03-07
feat: allow repeated face error if the repeated face is from same user

jowparks:
## Summary
When a user submits a KYC attempt, if they fail they will be flagged in subsequent transactions as duplicate face detected.

Here we check to make sure all duplicate faces that are detected are associated with users account associated with the current callback context.
## Testing Plan
Tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/redemptions/redemption.service.spec.ts
+++ b/src/redemptions/redemption.service.spec.ts
@@ -2,11 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { INestApplication } from '@nestjs/common';
-import { KycStatus } from '@prisma/client';
+import { DecisionStatus, KycStatus } from '@prisma/client';
 import assert from 'assert';
 import faker from 'faker';
 import { v4 as uuid } from 'uuid';
 import { AIRDROP_CONFIG } from '../common/constants';
+import { ImageChecksLabel } from '../jumio-api/interfaces/jumio-transaction-retrieve';
 import { WORKFLOW_RETRIEVE_FIXTURE } from '../jumio-kyc/fixtures/workflow';
 import { PrismaService } from '../prisma/prisma.service';
 import { bootstrapTestApp } from '../test/test-app';
@@ -37,9 +38,9 @@ describe('RedemptionServiceSpec', () => {
       jest.clearAllMocks();
     });
 
-    it('should return banned if user from PRK', () => {
+    it('should return banned if user from PRK', async () => {
       const fixture = WORKFLOW_RETRIEVE_FIXTURE('PROCESSED', 'PRK', 'PASSED');
-      const status = redemptionService.calculateStatus(fixture);
+      const status = await redemptionService.calculateStatus(fixture);
       expect(status).toEqual({
         status: KycStatus.FAILED,
         failureMessage: 'Failure: Banned Country',
@@ -53,9 +54,9 @@ describe('RedemptionServiceSpec', () => {
       });
     });
 
-    it('should not ban with acceptable country', () => {
+    it('should not ban with acceptable country', async () => {
       const fixture = WORKFLOW_RETRIEVE_FIXTURE('PROCESSED', 'CHL', 'PASSED');
-      const status = redemptionService.calculateStatus(fixture);
+      const status = await redemptionService.calculateStatus(fixture);
       expect(status).toEqual({
         status: KycStatus.SUBMITTED,
         failureMessage: null,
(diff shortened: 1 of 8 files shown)


src/redemptions/redemption.service.ts
```diff
@@ -138,6 +150,59 @@ export class RedemptionService {
     return null;
   }
```
jowparks, line 152: Added function

src/redemptions/redemption.service.ts
```diff
@@ -78,13 +82,21 @@ export class RedemptionService {
         idDetails,
       };
     }
+    const repeatedFace = await this.repeatedFaceCheck(transactionStatus);
```
lwisne, line 85: I'm still figuring out exact recommendations, but the duplication of the liveness/similarity/etc checks in this function and in `repeatedFaceCheck` strikes me as a code smell

src/redemptions/redemption.service.ts
```diff
@@ -138,6 +150,59 @@ export class RedemptionService {
...
+      response.capabilities.liveness.map((i) => i.decision.type),
+      response.capabilities.similarity.map((i) => i.decision.type),
+      response.capabilities.usability.map((i) => i.decision.type),
+    );
+    // if any other check failed, we can't pass
+    for (const potentialFailure of potentialFailures) {
+      if (potentialFailure !== 'PASSED') {
+        return null;
+      }
+    }
```
lwisne, line 170: I think you can delete everything above this point, and just continue execution of the current flow instead of early return

src/redemptions/redemption.service.ts
```diff
@@ -138,6 +150,59 @@ export class RedemptionService {
...
+    // for imageChecks, only duplicate face should pass
+    const failedImageChecks = response.capabilities.imageChecks.filter(
+      (i) => i.decision.type !== 'PASSED',
```
lwisne, line 173: is there another status besided `PASSED`? Should we be precise with the type we know accompanies a duplicated face?

src/redemptions/redemption.service.ts
```diff
@@ -138,6 +150,59 @@ export class RedemptionService {
...
+      const currentUserId = Number(response.workflow.customerInternalReference);
+      if (foundWorkflow.user_id !== currentUserId) {
+        return {
+          status: KycStatus.FAILED,
+          failureMessage: `User with multiple accounts detected (previous user id=${foundWorkflow.user_id} , current user id=${currentUserId}), this transaction matches face in transaction_id: ${foundWorkflow.workflow_execution_id}`,
+        };
+      }
+    }
+    return {
+      status: KycStatus.SUBMITTED,
```
lwisne, line 200: I woud try to avoid KycStatus at all here - have this be more encapsulated. Input a jumio transaction, and output is repeated face data from that tx.

Then the outer function pulls this data and uses it accordingly.

lwisne (changes requested):
Left some comments - I would like to see more encapsulation of the repeatedFace functionality.

Also, there is a world in which we use the similarityScore for this somehow, right? Looks like that's just a passthrough, but it seems like this would logically belong.

src/redemptions/redemption.service.ts
jowparks: decision.type is `WARNING` for `REPEATED_FACE` label, there is another label for warning  `DIFFERENT_PERSON`, so we should check them all (ie not all `WARNING`s are `REPEATED_FACE`

src/redemptions/redemption.service.ts
```diff
@@ -138,6 +150,59 @@ export class RedemptionService {
...
+      }
+    }
+    // for imageChecks, only duplicate face should pass
+    const failedImageChecks = response.capabilities.imageChecks.filter(
+      (i) => i.decision.type !== 'PASSED',
+    );
+    let duplicateFaceWorkflows: string[] = [];
+    for (const imageCheck of failedImageChecks) {
+      if (imageCheck.decision.details.label !== 'REPEATED_FACE') {
+        return null;
```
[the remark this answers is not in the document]
jowparks: I see what you are saying, going to refactor the code a bit based on your other suggestions and see if this comment still matters. I don't think it will :)

src/redemptions/redemption.service.ts
jowparks: I like this, I will need to split the function into two, because I want to check for intentional pass (two transactions same user) and hard fail (scammer case).

src/redemptions/redemption.service.ts
```diff
@@ -61,80 +76,106 @@ export class RedemptionService {
...
+    const repeatedFaceWorkflowIds = this.getRepeatedFaceWorkflowIds(
+      response.capabilities.imageChecks,
+    );
+    const multiAccountFailure = await this.multiAccountFailure(
+      userId,
+      repeatedFaceWorkflowIds,
+    );
+
     // TODO: HANDLE WARN, use decision.risk.score?
-    const failure =
```
jowparks, line 82: Consolidated these into a single check since I had to pull out labels anyways for the call to `this.hasOnlyDuplicateFaceFailures(labels)`

src/redemptions/redemption.service.spec.ts
```diff
@@ -180,4 +181,124 @@ describe('RedemptionServiceSpec', () => {
       );
     });
   });
+
```
jowparks, line 184: TODO: tomorrow fix tests

src/redemptions/redemption.service.ts
```diff
@@ -78,63 +87,99 @@ export class RedemptionService {
...
-      this.extractionStatus(transactionStatus) ||
-      this.usabilityStatus(transactionStatus);
+    const failure = multiAccountFailure || this.labelFailure(labels);
     if (failure) {
       return {
         status: KycStatus.FAILED,
         failureMessage: failure,
         idDetails,
       };
     }
```
jowparks, line 106: Even though we technically don't need `!multiAccountFailure` currently, if someone moved this code block higher up in the function, this would unexpectedly pass

src/redemptions/redemption.service.ts
```diff
@@ -78,63 +87,99 @@ export class RedemptionService {
...
-
+    if (!multiAccountFailure && this.hasOnlyDuplicateFaceFailures(labels)) {
```
lwisne, line 107: think the `!multiAccountFailure` is spurious due to preceding block

jowparks: See this comment:
https://github.com/iron-fish/ironfish-api/pull/1323#discussion_r1128222672

lwisne (approved):

lwisne:
appreciate all the new tests!

src/redemptions/redemption.service.ts
lwisne: fair

PR #1325, opened 2023-03-07
feat: add allocations table

jowparks:
## Summary
Adds allocations table, I am planning on calculating this via graphile job that will be triggered by protected endpoint. If the allocations exist, I will throw error and require user to wipe table manually if they want to regenerate.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -77,6 +77,10 @@ model Redemption {
   public_address   String    @db.VarChar
   id_details       Json?
   failure_message  String?   @db.VarChar
+  pool1_ore        BigInt?
+  pool2_ore        BigInt?
+  pool3_ore        BigInt?
+  pool4_ore        BigInt?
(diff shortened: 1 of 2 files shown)


prisma/schema.prisma
```diff
@@ -368,6 +369,21 @@ model AssetDescription {
   @@map("asset_descriptions")
 }
```
jowparks, line 371: Let me know if we need any other fields @lwisne

prisma/schema.prisma
```diff
@@ -412,3 +428,10 @@ enum KycStatus {
   SUBMITTED
   SUCCESS
 }
+
+enum AllocationPool {
```
jowparks, line 432: We could make this lowercase to match the config here:
https://github.com/iron-fish/ironfish-api/blob/d8d32b077565cc5ac497b61da3d30bdd0b0a99a2/src/common/constants.ts#L69
But this will not match convention for enum. I am fine either way, we can do case conversion

PR #1327, opened 2023-03-07
feat: add ip address to redemption model

jowparks:
## Summary
Captures IP address of client at time of redemption creation (first kyc attempt)
## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```


prisma/migrations/20230307211548_add_ip_address_to_redemption/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  Warnings:
+
+  - Added the required column `ip_address` to the `redemptions` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "redemptions" ADD COLUMN     "ip_address" VARCHAR NOT NULL;
```
rohanjadvani, line 8: Does this need to be backfilled?

```
ironfish-api-production::DATABASE=> SELECT COUNT(*) FROM redemptions;
 count
-------
    20
(1 row)
```

src/jumio-kyc/kyc.controller.spec.ts
```diff
@@ -149,6 +149,18 @@ describe('KycController', () => {
 
       expect(body).toMatchObject({ can_attempt: false });
     });
+    it('ip address is saved', async () => {
```
rohanjadvani, line 152: nit: new lines between tests

rohanjadvani (changes requested):
We need to make this new column nullable. Otherwise the migration will fail and the change won't deploy properly.

prisma/migrations/20230307211548_add_ip_address_to_redemption/migration.sql
jowparks: Lets make it nullable, I think this is a better approach. Honestly I am a bit surprised the prisma migration system didn't pick up on this?

rohanjadvani: Prisma adds a warning to the top of the file. I don't think the ORM can raise anything beyond that unless you point it at a db.

jowparks: I would have expected it to error out like Django with something like "Creating a nonnullable column on a model that already exists, either provide default value or make column nullable" 🤷

rohanjadvani (approved):

PR #1333, opened 2023-03-09
feat: sanctions screening response updates (1/3)

jowparks:
## Summary
Sanction screening for inline Jumio workflow KYC/Screening (workflow 10032).

## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-kyc/fixtures/workflow.ts
+++ b/src/jumio-kyc/fixtures/workflow.ts
@@ -5,6 +5,7 @@ import { DecisionStatus } from '@prisma/client';
 import {
   ImageCheck,
   JumioTransactionRetrieveResponse,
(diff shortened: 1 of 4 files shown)


src/jumio-kyc/fixtures/workflow.ts
```diff
@@ -14,6 +15,7 @@ export const WORKFLOW_RETRIEVE_FIXTURE = (
   decisionStatus: DecisionStatus,
   userId = '1',
   imageCheck: ImageCheck = IMAGE_CHECK_FIXTURE,
```
jowparks, line 17: In another PR I am gonna make this input an object instead of parameters, it is getting too big.

PR #1334, opened 2023-03-09
feat: add new sanction screening manual trigger endpoint (3/3)

jowparks:
## Summary
1. Adds endpoint for manually triggering standalone sanction screening workflows by userId.
2. Adds a method to create/submit standalone screening
   a. POST: Create new workflow (10010) with existing `accountId`
   b. GET: Retrieve PII from Jumio hosted workflow
   b. POST:  push search data (name, dob, country) to workflow from 2a
   c. PUT: force search initiation 
4. Callback will optionally run `redemption.calculateStandaloneSanctionStatus` if the workflow for the callback was `10010`

## Testing Plan
tests pass
NOTE: I tested this via true integration locally and it worked.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-api/jumio-api.service.ts
+++ b/src/jumio-api/jumio-api.service.ts
@@ -5,8 +5,11 @@ import { BadRequestException, Injectable } from '@nestjs/common';
 import axios, { AxiosRequestConfig } from 'axios';
 import crypto from 'crypto';
 import { ApiConfigService } from '../api-config/api-config.service';
+import { ScreeningData } from '../jumio-kyc/interfaces/screening-data';
 import { LoggerService } from '../logger/logger.service';
 import { JumioAccountCreateResponse } from './interfaces/jumio-account-create';
+import { JumioPutStandaloneScreening } from './interfaces/jumio-put-standalone-screening';
+import { JumioStandaloneUpload } from './interfaces/jumio-standalone-screening-create';
 import { JumioTransactionRetrieveResponse } from './interfaces/jumio-transaction-retrieve';
 
 @Injectable()
@@ -19,6 +22,7 @@ export class JumioApiService {
   async transactionStatus(
     jumio_account_id: string,
     jumio_workflow_execution_id: string,
+    sanitize = true,
   ): Promise<JumioTransactionRetrieveResponse> {
     const baseUrl = this.config.get<string>('JUMIO_URL');
     const url = `https://retrieval.${baseUrl}/accounts/${jumio_account_id}/workflow-executions/${jumio_workflow_execution_id}`;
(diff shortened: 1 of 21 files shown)


src/jumio-kyc/kyc.service.ts
```diff
@@ -201,10 +201,7 @@ export class KycService {
     transaction: JumioTransaction,
   ): Promise<void> {
     // Don't update redemption anymore for a user that has already passed KYC
```
jowparks, line 203: We do need to update if we do sanctions and want to go from SUBMITTED->SUCCESS

src/jumio-transactions/jumio-transaction.service.ts
```diff
@@ -58,9 +58,10 @@ export class JumioTransactionService {
   async findByWorkflowExecutionId(
     workflowExecutionId: string,
   ): Promise<JumioTransaction | null> {
-    return await this.prisma.jumioTransaction.findFirst({
+    const workflows = await this.prisma.jumioTransaction.findMany({
       where: { workflow_execution_id: workflowExecutionId },
     });
```
jowparks, line 63: We should use latest one if we somehow end up with duplicates (although this should never happen)

src/jumio-kyc/kyc.service.ts
```diff
@@ -194,15 +199,65 @@ export class KycService {
     });
   }
```
jowparks, line 201: Main code

src/jumio-api/jumio-api.service.ts
```diff
@@ -100,6 +107,98 @@ export class JumioApiService {
...
+      if (axios.isAxiosError(error)) {
+        this.logger.warn(`${JSON.stringify(error.response?.data)}`);
+
+        throw new BadRequestException(`Error updating standalone ${url}`);
+      }
+      throw error;
+    });
+
+    return response.data as JumioPutStandaloneScreening;
+  }
```
jowparks, line 164: If I used the code below (`axios.put<T>`) instead of the `axios(config)` method I used above, I was getting an error for invalid token when doing integration testing, not sure why.
```
{"level":40,"time":1678503280868,"pid":34131,"hostname":"Iron-Fish-MBP.local","msg":"{\"type\":\"about:blank\",\"title\":\"Unauthorized: wrong authorization token\",\"status\":401,\"detail\":\"[]\",\"instance\":\"/error\"}"}
```
```
  async putStandaloneScreening(
    url: string,
    token: string,
  ): Promise<JumioPutStandaloneScreening> {
    const config = {
      method: 'put',
      url: url,
      headers: {
        'User-Agent': 'IronFish Website/v1.0',
        'Content-Type': 'application/json',
        Authorization: `Bearer ${token}`,
      },
    };
    const promise = axios(config);

    const response = await promise.catch((error) => {
      if (axios.isAxiosError(error)) {
        this.logger.warn(`${JSON.stringify(error.response?.data)}`);

        throw new BadRequestException(`Error updating standalone ${url}`);
      }
      throw error;
    });

    return response.data as JumioPutStandaloneScreening;
  }
  
```

```
  async putStandaloneScreening(
    url: string,
    token: string,
  ): Promise<JumioPutStandaloneScreening> {
    const config = {
      headers: {
        'User-Agent': 'IronFish Website/v1.0',
        'Content-Type': 'application/json',
        Authorization: `Bearer ${token}`,
      },
    };
    const promise = axios.put<JumioPutStandaloneScreening>(url, config);

    const response = await promise.catch((error) => {
      if (axios.isAxiosError(error)) {
        this.logger.warn(`${JSON.stringify(error.response?.data)}`);

        throw new BadRequestException(`Error updating standalone ${url}`);
      }
      throw error;
    });

    return response.data;
}
```

src/jumio-transactions/jumio-transaction.service.spec.ts
```diff
@@ -84,56 +82,4 @@ describe('JumioTransactionService', () => {
       });
     });
   });
-
```
jowparks, line 87: I didn't realize you had added santizing already, deleted my version of it.

NullSoldier (approved):

PR #1336, opened 2023-03-09
feat: watchlist screening success/failure integration

jowparks:
## Summary
1. If the user has `searchResults!==0` returned OR jumio has an `ALERT` label on one of the watchlist entries we fail the user indefinitely for KYC.
2. If the whole workflow is successful now, we can move user to `SUCCESS` label rather than `SUBMITTED`
## Testing Plan
Tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-api/interfaces/jumio-transaction-retrieve.ts
+++ b/src/jumio-api/interfaces/jumio-transaction-retrieve.ts
@@ -87,6 +87,17 @@ export type ImageChecksLabel =
   | 'PRECONDITION_NOT_FULFILLED'
   | 'TECHNICAL_ERROR';
 
+export type WatchlistScreeningLabels =
+  | 'NOT_ENOUGH_DATA'
+  | 'VALIDATION_FAILED'
+  | 'INVALID_MERCHANT_SETTINGS'
+  | 'TECHNICAL_ERROR'
+  | 'EXTRACTION_NOT_DONE'
+  | 'NO_VALID_ID_CREDENTIAL'
+  | 'PRECONDITION_NOT_FULFILLED'
+  | 'OK'
+  | 'ALERT';
+
 export type ImageCheck = {
   id: string;
   credentials: {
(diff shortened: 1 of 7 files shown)


src/jumio-api/interfaces/jumio-transaction-retrieve.ts
```diff
@@ -102,9 +113,31 @@ export type ImageCheck = {
   data: {
     faceSearchFindings: {
       status: string;
```
jowparks, line 115: I discovered this is an optional return field in their API, it isn't present if there are no findings, but is present and is a list if there is.

jowparks: (this is undocumented in their API)

src/jumio-api/interfaces/jumio-transaction-retrieve.ts
```diff
@@ -128,6 +161,7 @@ export interface JumioTransactionRetrieveResponse {
   workflow: {
     id: string;
     definitionKey: string;
```
jowparks, line 163: We were missing this field in our type

src/redemptions/redemption.service.ts
```diff
@@ -84,11 +85,14 @@ export class RedemptionService {
     }
     if (transactionStatus.decision.type === DecisionStatus.PASSED) {
       return {
```
jowparks, line 87: We can now succeed since watchlist is integrated!

src/jumio-api/interfaces/jumio-transaction-retrieve.ts
```diff
@@ -87,6 +87,17 @@ export type ImageChecksLabel =
   | 'PRECONDITION_NOT_FULFILLED'
   | 'TECHNICAL_ERROR';
```
jowparks, line 89: Documentation here:
https://jumio.github.io/kyx/integration-guide.html#capabilitieswatchlistscreening

src/jumio-api/interfaces/jumio-transaction-retrieve.ts
```diff
@@ -192,7 +226,7 @@ export interface JumioTransactionRetrieveResponse {
       id: string;
       credentials: [
         {
```
jowparks, line 228: There were 1. incorrect type value, 2. undocumented undefined fields in the API response

mat-if (approved):

PR #1354, opened 2023-03-13
feat: allow liveness warnings to pass on low risk KYC

jowparks:
## Summary
We have a list of labels that are considered passable when there are no other failures present.
1. Duplicate face will occur if user has previous KYC attempt that has failed (even for same account).
2. Liveness undetermined is an added check to make sure photograph was not use for face detection, if it can't complete successfully it will give status: `LIVENESS_UNDETERMINED`, which is non-block per Jumio reps. We were told below `risk` of 50, we can disregard this warning.
## Testing Plan
Tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-kyc/fixtures/workflow.ts
+++ b/src/jumio-kyc/fixtures/workflow.ts
@@ -5,9 +5,11 @@ import { DecisionStatus } from '@prisma/client';
 import {
   ImageCheck,
   JumioTransactionRetrieveResponse,
+  LivenessCheck,
   WatchlistScreenCheck,
 } from '../../jumio-api/interfaces/jumio-transaction-retrieve';
 import { IMAGE_CHECK_FIXTURE } from './image-check';
+import { LIVENESS_CHECK_FIXTURE } from './liveness-check';
 import { WATCHLIST_SCREEN_FIXTURE } from './watch-list';
 
 export const WORKFLOW_RETRIEVE_FIXTURE = (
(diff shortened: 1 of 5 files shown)


src/jumio-kyc/fixtures/workflow.ts
```diff
@@ -5,22 +5,26 @@ import { DecisionStatus } from '@prisma/client';
...
   ImageCheck,
   JumioTransactionRetrieveResponse,
+  LivenessCheck,
   WatchlistScreenCheck,
 } from '../../jumio-api/interfaces/jumio-transaction-retrieve';
 import { IMAGE_CHECK_FIXTURE } from './image-check';
+import { LIVENESS_CHECK_FIXTURE } from './liveness-check';
 import { WATCHLIST_SCREEN_FIXTURE } from './watch-list';
 
 export const WORKFLOW_RETRIEVE_FIXTURE = (
```
jowparks, line 15: Putting in follow up PR to turn this constructor params into an object so that the function calls aren't bloated with `undefined`s

src/redemptions/redemption.service.spec.ts
```diff
@@ -341,10 +343,53 @@ describe('RedemptionServiceSpec', () => {
...
       expect(acceptableFace).toBe(true);
     });
+
+    it('should allow LIVENESS_UNDETERMINED warning for success if risk score acceptable', async () => {
+      const fixture = WORKFLOW_RETRIEVE_FIXTURE(
+        'PROCESSED',
+        'CHL',
+        DecisionStatus.WARNING,
+        '1231232',
+        IMAGE_CHECK_FIXTURE('OK'),
```
jowparks, line 356: Putting in follow up PR to turn this constructor params into an object so that the function calls aren't bloated with undefineds

src/redemptions/redemption.service.ts
```diff
@@ -112,10 +112,14 @@ export class RedemptionService {
         idDetails,
       };
     }
```
jowparks, line 114: Main changes are here

dgca (approved):

PR #1356, opened 2023-03-14
feat: make KYC unsuccessful if age below 18

jowparks:
## Summary
Two things:
1. On callback, KYC attempt is marked `TRY_AGAIN` with failure message telling user is below age requirement
2. On eligibility check, users can be eligible to attempt if age is below requirement, but warned their parents must be custodian (will be set to `TRY_AGAIN` if kyc is attempted).

Backfill script (very slow response, no need to rate limit):
```python
import requests

tsv = 'passed_users.tsv'
out = 'output.txt'
token = ''

def retrieve_age():
    base_url = "https://retrieval.amer-1.jumio.ai/api/v1/accounts/{}/workflow-executions/{}"
    headers = {
        'Accept': 'application/json',
        'Content-Type': 'application/json',
        'User-Agent': 'IronFish Website/v1.0',
        'Authorization': f'Basic {token}'
    }

    num = 0

    data = []
    with open(tsv) as f:
        for row in f.readlines():
            print(row.split())
            data.append(row.split())

    with open(out, 'a') as f:
        for account_id, workflow_execution_id, user_id in data:
            print(f'processing {num}')
            url = base_url.format(account_id, workflow_execution_id)
            response = requests.get(url, headers=headers)
            if response.status_code == 200:
                ages = [int(e['data'].get('currentAge', -1)) for e in response.json()['capabilities']['extraction']]
                age = min(ages)
                if (age < 18):
                    f.write(f'UPDATE redemptions set kyc_status = TRY_AGAIN, age = {age} where user_id = {user_id}')
                else:
                    print(f'Age: {age}')
            else:
                print("Error: Could not retrieve credentials for account_id {} and workflow_execution_id {}".format(account_id, workflow_execution_id))
            num += 1


retrieve_age()
```
## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-kyc/fixtures/workflow.ts
+++ b/src/jumio-kyc/fixtures/workflow.ts
@@ -3,43 +3,41 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { DecisionStatus } from '@prisma/client';
 import {
+  ExtractionCheck,
   ImageCheck,
   JumioTransactionRetrieveResponse,
   LivenessCheck,
   WatchlistScreenCheck,
   WorkflowStatus,
 } from '../../jumio-api/interfaces/jumio-transaction-retrieve';
+import { EXTRACTION_CHECK_FIXTURE } from './extraction-check';
 import { IMAGE_CHECK_FIXTURE } from './image-check';
 import { LIVENESS_CHECK_FIXTURE } from './liveness-check';
 import { WATCHLIST_SCREEN_FIXTURE } from './watch-list';
 
 export type JumioWorkflowRetrieveParams = {
   workflowStatus?: WorkflowStatus;
-  idCountryCode?: string;
   workflowId?: string;
   accountId?: string;
-  firstName?: string;
-  lastName?: string;
   decisionStatus?: DecisionStatus;
   userId?: string;
   imageCheck?: ImageCheck;
   watchlistCheck?: WatchlistScreenCheck;
   livenessCheck?: LivenessCheck;
+  extractionCheck?: ExtractionCheck;
   riskScore?: number;
 };
 export const WORKFLOW_RETRIEVE_FIXTURE = ({
   workflowStatus = 'PROCESSED',
-  idCountryCode = 'CHL',
   decisionStatus = DecisionStatus.PASSED,
-  userId = '1',
-  imageCheck = IMAGE_CHECK_FIXTURE(),
-  watchlistCheck = WATCHLIST_SCREEN_FIXTURE(),
+  riskScore = 50,
   workflowId = 'eb776622-26a2-499b-b824-6449c92d1617',
   accountId = '684fcf21-bcf3-4916-b007-35dc70102f56',
-  firstName = 'Jason',
-  lastName = 'Spafford',
+  userId = '1',
+  extractionCheck = EXTRACTION_CHECK_FIXTURE(),
+  imageCheck = IMAGE_CHECK_FIXTURE(),
+  watchlistCheck = WATCHLIST_SCREEN_FIXTURE(),
   livenessCheck = LIVENESS_CHECK_FIXTURE(),
-  riskScore = 50,
 }: JumioWorkflowRetrieveParams = {}): JumioTransactionRetrieveResponse => {
   return {
     workflow: {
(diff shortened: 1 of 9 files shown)


src/jumio-kyc/kyc.service.ts
```diff
@@ -274,13 +274,13 @@ export class KycService {
         : this.redemptionService.calculateStandaloneWatchlistStatus(status);
 
     // Has our user's KYC status changed
```
jowparks, line 276: We now need to update even if user status goes from `TRY_AGAIN` -> `TRY_AGAIN` since this could be caused by age issue.

src/redemptions/redemption.service.ts
```diff
@@ -255,6 +279,7 @@ export class RedemptionService {
         id_details: instanceToPlain(data.idDetails),
         failure_message: data.failureMessage,
         public_address: data.publicAddress,
```
jowparks, line 281: If add age as param that is `undefined` prisma trys to use `NaN` and fails. Must omit property if not defined

src/redemptions/redemption.service.ts
```diff
@@ -430,4 +462,7 @@ export class RedemptionService {
 
     return { attemptable: true, reason: '' };
   }
+
```
jowparks, line 465: https://iron-fish.slack.com/archives/C04MQM3P2DT/p1678814001893989?thread_ts=1678749618.280909&cid=C04MQM3P2DT

src/redemptions/redemption.service.ts
```diff
@@ -430,4 +462,7 @@ export class RedemptionService {
...
+  minorAgeMessage = (age: number): string =>
```
jowparks, line 466: @danield9tqh this message will get piped into `can_attempt_reason` and `can_attempt=true`

src/redemptions/redemption.service.ts
```diff
@@ -71,9 +78,17 @@ export class RedemptionService {
         status: KycStatus.FAILED,
         failureMessage: 'Failure: Banned Country',
         idDetails,
+        age,
+      };
+    }
```
jowparks, line 83: if `undefined`, will bypass this check which is expected

src/redemptions/redemption.service.ts
dguenther: This seems odd... maybe a prisma bug?

dguenther (approved):

src/jumio-kyc/fixtures/workflow.ts
```diff
@@ -153,7 +155,7 @@ export const WORKFLOW_RETRIEVE_FIXTURE = ({
             documentNumber: '1111111',
             optionalMrzField1: 'Z11',
             optionalMrzField2: '11111111',
-            currentAge: '70',
+            currentAge: age,
```
NullSoldier, line 158: is this provided on every document or is it optional?

src/jumio-kyc/kyc.service.ts
jowparks: If the user fails because their age < 18, the return status will be `TRY_AGAIN`. If they were already in `TRY_AGAIN` status, without this change we wouldn't persist their age.

I believe this guard was to prevent reprocessing of callbacks.

src/redemptions/redemption.service.ts
```diff
@@ -430,4 +462,7 @@ export class RedemptionService {
 
     return { attemptable: true, reason: '' };
   }
+
+  minorAgeMessage = (age: number): string =>
+    `The age for your KYC (${age} years old) is lower than the legal requirement of 18. Please give custodianship to your parent/guardian and have them KYC.`;
```
NullSoldier, line 467: I prefer that we write simple english. I can immedietly tell that an engineer wrote this message.

src/redemptions/redemption.service.ts
jowparks: Yea I am guessing thats the case, I saw some open tickets in prisma relative to this, just wrote it this way as workaround.
The exact error was as follows (when age was undefined)

```
→ 275   return client.redemption.update({
          data: {
            kyc_status: 'SUCCESS',
            jumio_account_id: undefined,
            id_details: [
              {
                id_issuing_country: 'CHL',
                id_subtype: undefined,
                id_type: undefined
              }
            ],
            failure_message: undefined,
            public_address: undefined,
            age: NaN
                 ~~~
          },
          where: {
            id: 7
          }
        })

Argument age: Got invalid value NaN on prisma.updateOneRedemption. Provided Float, expected Int or NullableIntFieldUpdateOperationsInput or Null.
```
Guessing it is interpreting undefined as `NaN` since the parameter is a number and Postgres supports NaN for some numeric fields

This occurs when I had the code as:
```
  async update(
    redemption: Redemption,
    data: {
      kycStatus?: KycStatus;
      jumioAccountId?: string;
      idDetails?: IdDetails[];
      failureMessage?: string;
      publicAddress?: string;
      age?: number;
    },
    prisma?: BasePrismaClient,
  ): Promise<Redemption> {
    const client = prisma ?? this.prisma;

    return client.redemption.update({
      data: {
        kyc_status: data.kycStatus,
        jumio_account_id: data.jumioAccountId,
        id_details: instanceToPlain(data.idDetails),
        failure_message: data.failureMessage,
        public_address: data.publicAddress,
        age: data.age,
      },
      where: {
        id: redemption.id,
      },
    });
  }
```
FWIW prisma handles undefined properly for string fields 🤷

src/redemptions/redemption.service.ts
```diff
@@ -255,6 +279,7 @@ export class RedemptionService {
         id_details: instanceToPlain(data.idDetails),
         failure_message: data.failureMessage,
         public_address: data.publicAddress,
+        ...(data.age ? { age: data.age } : {}),
```
NullSoldier, line 282: Aren't undefined filtered out by the prisma filter?

```suggestion
        data: data.age
```

src/jumio-kyc/fixtures/workflow.ts
jowparks: API says every document, but there are other fields that haven't been there when it was a nonoptional return, I will guard for this case.

src/redemptions/redemption.service.ts
jowparks: https://github.com/iron-fish/ironfish-api/pull/1356#discussion_r1136147718

PR #1359, opened 2023-03-16
chore: reset db at start of each test file

jowparks:
## Summary
DB data shouldn't bleed between test suites, this makes sure that doesn't happen.

## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/health/health.controller.spec.ts
+++ b/src/health/health.controller.spec.ts
@@ -39,17 +39,4 @@ describe('HealthController', () => {
       });
     });
   });
-
-  describe('GET /health/deposit', () => {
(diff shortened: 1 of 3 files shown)


src/health/health.controller.spec.ts
```diff
@@ -39,17 +39,4 @@ describe('HealthController', () => {
       });
     });
   });
-
```
jowparks, line 42: This test is failing because there is no setup for the test, it was rely on other files.

rohanjadvani (approved):

PR #1369, opened 2023-03-20
Add error handling in Auth

ygao76:
## Summary
Throw detailed error message during authentication.

## Testing Plan
spec test

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/auth/guards/magic-link.guard.ts
+++ b/src/auth/guards/magic-link.guard.ts
@@ -1,8 +1,42 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { Injectable } from '@nestjs/common';
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import {
+  BadRequestException,
+  Injectable,
+  NotFoundException,
+  UnauthorizedException,
+} from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
 
 @Injectable()
-export class MagicLinkGuard extends AuthGuard('magic-link') {}
+export class MagicLinkGuard extends AuthGuard('magic-link') {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-explicit-any
+  handleRequest(err: any, user: any, info: any, context: any, status: any) {
+    if (info && !user) {
+      switch (info) {
+        case 'no_account_found':
+          throw new NotFoundException(
+            'Error: No Iron Fish account exists for this email.',
+          );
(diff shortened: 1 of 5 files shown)


src/auth/guards/magic-link.guard.ts
```diff
@@ -1,8 +1,42 @@
...
+  NotFoundException,
+  UnauthorizedException,
+} from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
 
 @Injectable()
-export class MagicLinkGuard extends AuthGuard('magic-link') {}
+export class MagicLinkGuard extends AuthGuard('magic-link') {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-explicit-any
+  handleRequest(err: any, user: any, info: any, context: any, status: any) {
```
jowparks, line 18: It is lame the passport library doesn't have better types than this, but nothing you can do 🤷

jowparks (approved):
LGTM, just verify login is working as expected in staging before using the github action to promote to prd.

src/auth/guards/magic-link.guard.ts
rohanjadvani: `context` should be `ExecutionContext` and `status` is optional. We can update this method to make it stricter

rohanjadvani: You can still override types for `err`, `user` (this needs to be parameterized), and `info` because they are `any`. TS class extension will allow for that

src/auth/guards/magic-link.guard.ts
```diff
@@ -1,8 +1,42 @@
...
+    if (info && !user) {
+      switch (info) {
+        case 'no_account_found':
```
rohanjadvani, line 21: In other places in the API (for example [here](https://github.com/iron-fish/ironfish-api/blob/master/src/faucet-transactions/faucet-transactions.service.ts#L55-L58)), we return a `code` and a `message`. It's cleaner to do that if you need to add more error codes in the future, and you can return a single `UnauthorizedException` here.

src/auth/guards/magic-link.guard.ts
```diff
@@ -1,8 +1,42 @@
...
+          throw new NotFoundException(
+            'Error: No Iron Fish account exists for this email.',
+          );
+        case 'empty_email':
+          throw new BadRequestException(
```
rohanjadvani, line 26: We generally use `UnprocessableEntityException` throughout the API. However, most auth guards in NestJS return unauthorized, so i think that'd be fine too.

src/auth/guards/magic-link.guard.ts
```diff
@@ -1,8 +1,42 @@
...
+          throw new UnauthorizedException('Error: Login is disabled.');
+        default:
+          throw new UnauthorizedException(info);
+      }
+    }
+
+    if (err || !user) {
+      throw err || new UnauthorizedException();
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
rohanjadvani, line 39: We can remove this warning if we clean up above suggestions

PR #1374, opened 2023-03-21
Surface failureReason and failureUrl

NullSoldier:
## Summary

This surfaces the calculateStatus reason for the last jumio transaction.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-kyc/kyc.controller.ts
+++ b/src/jumio-kyc/kyc.controller.ts
@@ -27,6 +27,7 @@ import { IntIsSafeForPrismaPipe } from '../common/pipes/int-is-safe-for-prisma.p
 import { fetchIpAddressFromRequest } from '../common/utils/request';
 import { GraphileWorkerPattern } from '../graphile-worker/enums/graphile-worker-pattern';
 import { GraphileWorkerService } from '../graphile-worker/graphile-worker.service';
+import { JumioTransactionRetrieveResponse } from '../jumio-api/interfaces/jumio-transaction-retrieve';
 import { JumioTransactionService } from '../jumio-transactions/jumio-transaction.service';
 import { RedemptionService } from '../redemptions/redemption.service';
 import { UsersService } from '../users/users.service';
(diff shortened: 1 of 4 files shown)


src/redemptions/redemption.service.spec.ts
```diff
@@ -139,133 +142,6 @@ describe('RedemptionServiceSpec', () => {
       );
     });
```
jowparks, line 144: Instead of deleting these tests, can you just update them to call `calculateStatus` instead of `isEligible` and update assertions?

src/redemptions/redemption.service.ts
```diff
@@ -446,48 +449,6 @@ export class RedemptionService {
       };
     }
```
jowparks, line 451: very nice, I like

src/jumio-kyc/kyc.controller.ts
```diff
@@ -151,14 +151,29 @@ export class KycController {
     const jumioTransaction =
       await this.jumioTransactionService.findLatestOrThrow(user);
 
+    // If we have a latest attempt, override with that reasons failure
+    if (jumioTransaction.last_workflow_fetch && !eligibility.reason) {
+      const { failureUrl, failureMessage } =
+        await this.redemptionService.calculateStatus(
+          jumioTransaction.last_workflow_fetch as unknown as JumioTransactionRetrieveResponse,
+        );
+
```
jowparks, line 160: This should work ok, but it could get hairy if this conditional is tweaked int he future. Currently it does truthy checks on a `string | null` type that is at times `''`, maybe you can just change the `HELP_URLS` to be `= null` instead of `= ''`

jowparks (approved):

src/redemptions/redemption.service.spec.ts
NullSoldier: There are already tests for calculateStatus, these duplicated those tests.

PR #1378, opened 2023-03-22
Refactor magic link guard

ygao76:
## Summary
Refactor magic link guard

## Testing Plan
spec

## Breaking Change

```
[ ] Yes
```
@rohanjadvani

### DIFF
--- a/src/auth/guards/magic-link.guard.ts
+++ b/src/auth/guards/magic-link.guard.ts
@@ -10,13 +10,14 @@ import {
   UnauthorizedException,
 } from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
+import passport from 'passport';
 
 @Injectable()
 export class MagicLinkGuard extends AuthGuard('magic-link') {
   handleRequest<User>(
     err: Error,
     user: User,
-    info: string,
+    info: passport.StrategyFailure,
     _context: ExecutionContext,
   ): User {
     if (info && !user) {
--- a/src/auth/guards/magic-link.guard.ts
+++ b/src/auth/guards/magic-link.guard.ts
@@ -17,8 +17,7 @@ export class MagicLinkGuard extends AuthGuard('magic-link') {
     err: Error,
     user: User,
     info: string,
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    context: ExecutionContext,
+    _context: ExecutionContext,
   ): User {
     if (info && !user) {
       throw new UnauthorizedException(info);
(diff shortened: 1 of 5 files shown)


src/auth/guards/magic-link.guard.ts
```diff
@@ -5,38 +5,28 @@
...
 @Injectable()
 export class MagicLinkGuard extends AuthGuard('magic-link') {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-explicit-any
-  handleRequest(err: any, user: any, info: any, context: any, status: any) {
+  handleRequest<User>(
+    err: Error,
+    user: User,
+    info: string,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    context: ExecutionContext,
```
rohanjadvani, line 21:
```suggestion
    _context: ExecutionContext,
```

You can do this to get rid of the warning

src/auth/strategies/magic-link.strategy.ts
```diff
@@ -47,7 +47,7 @@ export class MagicLinkStrategy extends PassportStrategy(
         user,
       } as MagicLinkContext;
     } catch {
-      return this.fail('no_account_found');
+      return this.fail('Error: No Iron Fish account exists for this email.');
```
rohanjadvani, line 50: I think it'd be nice for consistency to return objects here with a `code` and `message`, like we do [here](https://github.com/iron-fish/ironfish-api/blob/master/src/users/users-updater.ts#L30-L45). This could look like
```suggestion
      return this.fail({
        code: 'user_not_found',
        message: 'User not found for this email',
      });
```

What do you think?

rohanjadvani (commented):
Nice! This looks great. I left a couple minor suggestions to be more consistent with how we handle errors. It'll be good to merge and deploy after that.

src/auth/strategies/magic-link.strategy.ts
```diff
@@ -26,18 +26,27 @@ export class MagicLinkStrategy extends PassportStrategy(
   async authenticate(req: Request): Promise<void> {
     const { authorization } = req.headers;
     if (!authorization) {
-      return this.fail('unauthorized');
+      return this.fail({
+        code: 'no_authorization',
+        message: 'Error: Request header is invalid.',
```
rohanjadvani, line 31: nit: If we're checking `!authorization`, then the header is undefined or `''`

```suggestion
        message: 'Error: Request header is missing.',
```

rohanjadvani (changes requested):
One of the types is incorrect. We should fix that, and then I think we can merge it.

src/auth/strategies/magic-link.strategy.ts
ygao76: The request header is not missing, the request header is missing authorization data. That is why I put invalid request header. But I dont think we need to be specific to user about the authorization data is missing.

PR #1379, opened 2023-03-22
Improve auth email error handling

ygao76:
## Testing Plan
spec
## Breaking Change

```
[ ] Yes
```
@rohanjadvani

### DIFF
--- a/src/auth/strategies/magic-link.strategy.ts
+++ b/src/auth/strategies/magic-link.strategy.ts
@@ -42,10 +42,24 @@ export class MagicLinkStrategy extends PassportStrategy(
     let email;
     try {
       email = await this.magicLinkService.getEmailFromHeader(authorization);
-    } catch {
+    } catch (err) {
+      if (err instanceof Error) {
+        return this.fail({
+          code: err.name,
+          message: `Error: ${err.message}. Please try again.`,
+        });
+      }
+
       return this.fail({
         code: 'email_error',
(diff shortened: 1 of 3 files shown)


src/auth/strategies/magic-link.strategy.ts
```diff
@@ -36,10 +36,18 @@ export class MagicLinkStrategy extends PassportStrategy(
     let email;
     try {
       email = await this.magicLinkService.getEmailFromHeader(authorization);
-    } catch {
+    } catch (err) {
+      if (err instanceof Error) {
+        return this.fail(`${err.name}: ${err.message}`);
```
rohanjadvani, line 41: It feels weird here to fail with either error codes (that are interpreted by the guard) or strings. We should try to be consistent with what is passed to each `fail`

src/magic-link/magic-link.service.ts
```diff
@@ -20,7 +20,16 @@ export class MagicLinkService {
   }
 
   async getEmailFromToken(didToken: string): Promise<string> {
-    this.magic.token.validate(didToken);
+    try {
+      this.magic.token.validate(didToken);
```
rohanjadvani, line 24: Do we need to wrap this? Isn't there an additional error catch from the caller? If it's an error from Magic, we can probably just check codes from [here](https://github.com/magiclabs/magic-admin-js/blob/master/src/core/sdk-exceptions.ts)

PR #1390, opened 2023-03-28
Update deploy workflows for mainnet

danield9tqh:
## Summary
Getting rid of prod + staging apps for testnet archives so deploys will just happen directly to production now. Also renaming heroku apps to `testnet` + `mainnet`

## Breaking Change

```
[ ] Yes
```


.github/workflows/deploy-phase1.yml
```diff
@@ -1,19 +1,19 @@
...
-    name: Deploy Phase1 Staging API
+    name: Deploy Phase 1 API
     runs-on: ubuntu-latest
 
     steps:
       - name: Check out Git repository
         uses: actions/checkout@v2
       - uses: akhileshns/heroku-deploy@v3.12.12
         with:
-          heroku_app_name: ${{ secrets.HEROKU_APP_NAME_STAGING_PHASE1 }}
```
danield9tqh, line 17: Looking into documentation, makes no mention that app name should be secret and these secrets seem to provide minimal protection especially for archive apps https://github.com/marketplace/actions/deploy-to-heroku

dguenther (approved):
We'll probably need to do many fewer deploys after we mainnet launch, so I think this looks good and we can simplify if we need to

PR #1392, opened 2023-03-29
fix: repeated face same account

jowparks:
## Summary
There are failures resulting from someone attempting KYC > 5 times:

Example below is failing because he is marked as MISMATCHING_DATA_REPEATED_FACE, everything else succeeded. I checked all the linked transactions and they are all for the same account, once he finally used his passport Jumio then recognized all his previous attempts on the same account as a scammer
https://ironfish7143.zendesk.com/agent/tickets/91
https://customer-portal.netverify.com/app/#/transactions/aab161b4-51dc-4dd3-96e3-0eb61fb89464

### minor updates
- removed failure message from repeated face success
- removed check for `LIVENESS_UNDETERMINED` since it is caught by other logic now

## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/redemptions/redemption.service.ts
+++ b/src/redemptions/redemption.service.ts
@@ -279,6 +279,14 @@ export class RedemptionService {
           idDetails,
           age,
         };
+      } else if (this.hasOnlyBenignFaceWarnings(labels)) {
+        return {
+          status: KycStatus.SUCCESS,
+          failureUrl: null,
+          failureMessage: null,
+          idDetails,
+          age,
+        };
       }
     }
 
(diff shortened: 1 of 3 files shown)


src/redemptions/redemption.service.ts
```diff
@@ -312,18 +320,6 @@ export class RedemptionService {
       };
     }
 
-    if (
-      this.hasOnlyBenignWarnings(labels) &&
-      transactionStatus.decision.risk.score < 50
```
jowparks, line 317: Had to remove acceptable risk score to allow these passes. Note that scammers should be caught by line 273 in this file.

PR #1398, opened 2023-03-31
Add jwt token verification

ygao76:
## Summary
Add cookie checks to Magic Link strategy
## Testing Plan
spec
## Breaking Change

```
[ ] Yes
```
@rohanjadvani


package.json
```diff
@@ -82,6 +83,7 @@
     "class-validator": "0.13.2",
     "cluster": "0.7.7",
     "compression": "1.7.4",
+    "cookie": "0.5.0",
```
rohanjadvani, line 86: What do you think about using [`cookie-parser`](https://www.npmjs.com/package/cookie-parser)? We just need to add it as middleware in the server? It'll be a little cleaner.

src/auth/strategies/magic-link.strategy.ts
```diff
@@ -24,44 +26,87 @@ export class MagicLinkStrategy extends PassportStrategy(
...
+        const decoded = jwt.verify(token, this.config.get('JWT_TOKEN_SECRET'));
+
+        if (decoded.sub === undefined || decoded.sub.toString().length === 0) {
+          return this.fail({
+            code: 'invalid_token',
+            message: 'Error: Jwt token is missing email.',
+          });
+        }
+
+        email = decoded.sub.toString();
```
rohanjadvani, line 61: Is `sub` already a string?

src/auth/strategies/magic-link.strategy.ts
```diff
@@ -24,44 +26,87 @@ export class MagicLinkStrategy extends PassportStrategy(
...
+          code: 'email_error',
+          message: 'Error: Failed to get the email. Please try again.',
+        });
+      }
+    } else if (cookies) {
+      try {
+        const token: string = cookies.jwt;
+        const decoded = jwt.verify(token, this.config.get('JWT_TOKEN_SECRET'));
+
+        if (decoded.sub === undefined || decoded.sub.toString().length === 0) {
```
rohanjadvani, line 54: We can check `!decoded.sub`

src/auth/strategies/magic-link.strategy.ts
```diff
@@ -24,44 +26,87 @@ export class MagicLinkStrategy extends PassportStrategy(
...
+          return this.fail({
+            code: 'invalid_token',
+            message: 'Error: Jwt token is missing email.',
+          });
+        }
+
+        email = decoded.sub.toString();
+      } catch (err) {
+        if (err instanceof JsonWebTokenError) {
```
rohanjadvani, line 63: `JsonWebTokenError` has a `name` and `message`. Can we that message instead of wrapping it? I don't think we need to add anything more to that

src/users/me.controller.spec.ts
```diff
@@ -62,6 +64,38 @@ describe('MeController', () => {
...
+          countryCode: faker.address.countryCode('alpha-3'),
+        });
+
+        const token = jwt.sign({ sub: user.email, iat: Date.now() }, 'secret', {
+          expiresIn: '1d',
+        });
+
+        const { body } = await request(app.getHttpServer())
+          .get('/me')
+          .set(
```
rohanjadvani, line 81: We can use `.cookies`

ygao76: That method does not set the cookies. In the class file, it has this interface
`set(field: "Cookie", val: string[]): this;`

src/auth/strategies/magic-link.strategy.ts
ygao76: not really, it has an overloaded type.

src/auth/strategies/magic-link.strategy.ts
ygao76: Updated, I did not use it at first because the name in `JsonWebTokenError` is camel case but our code format is snake case.

src/test/test-app.ts
```diff
@@ -60,5 +61,6 @@ export async function bootstrapTestApp(): Promise<INestApplication> {
 
   const app = module.createNestApplication();
   app.use(json({ limit: '10mb' }));
+  app.use(cookieParser('secret'));
```
rohanjadvani, line 64: Should we use the config value here too?

src/auth/strategies/magic-link.strategy.ts
```diff
@@ -24,44 +25,74 @@ export class MagicLinkStrategy extends PassportStrategy(
...
+      } catch {
+        return this.fail({
+          code: 'email_error',
+          message: 'Error: Failed to get the email. Please try again.',
+        });
+      }
+    } else if (req.cookies) {
+      try {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+        const token: string = req.cookies.jwt;
```
rohanjadvani, line 50: We are using `ironfish_jwt` to be a bit more explicit

rohanjadvani (approved):

src/test/test-app.ts
ygao76: updated

PR #1399, opened 2023-03-31
feat(api): Add JWT login route

rohanjadvani:
## Summary

Add a login route which takes a JWT

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/auth/auth.controller.spec.ts
+++ b/src/auth/auth.controller.spec.ts
@@ -122,9 +122,9 @@ describe('AuthController', () => {
           .expect(HttpStatus.FOUND);
 
         expect(header.location).toBe(
-          `${config.get<string>(
-            'INCENTIVIZED_TESTNET_URL',
-          )}/leaderboard?toast=${btoa('User not found')}`,
+          `${config.get<string>('INCENTIVIZED_TESTNET_URL')}/login?toast=${btoa(
+            'User not found',
+          )}&persist=true`,
         );
       });
     });
@@ -155,9 +155,9 @@ describe('AuthController', () => {
           .expect(HttpStatus.FOUND);
 
         expect(header.location).toBe(
-          `${config.get<string>(
-            'INCENTIVIZED_TESTNET_URL',
-          )}/leaderboard?toast=${btoa('jwt expired')}`,
+          `${config.get<string>('INCENTIVIZED_TESTNET_URL')}/login?toast=${btoa(
+            'jwt expired',
+          )}&persist=true`,
         );
       });
     });
@@ -190,7 +190,7 @@ describe('AuthController', () => {
           .expect(HttpStatus.FOUND);
 
         expect(header.location).toBe(
-          `${config.get<string>('INCENTIVIZED_TESTNET_URL')}/leaderboard`,
+          `${config.get<string>('INCENTIVIZED_TESTNET_URL')}/login`,
         );
         expect(updateLastLoginAt).toHaveBeenCalledTimes(1);
       });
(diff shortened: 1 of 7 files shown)


src/auth/auth.controller.ts
```diff
@@ -54,4 +61,78 @@ export class AuthController {
...
+      }
+
+      throw new UnauthorizedException();
+    }
+  }
+
+  private setResponseHeaders(res: Response, token: string): void {
+    const cookieOptions: CookieOptions = {
+      httpOnly: true,
+      maxAge: MS_PER_DAY,
```
ygao76, line 125: Can you add domain value for more security?

src/auth/auth.controller.ts
```diff
@@ -54,4 +61,78 @@ export class AuthController {
...
+      throw new UnauthorizedException();
+    }
+
+    const email = this.verifyEmailFromJwt(token);
+
+    const user = await this.usersService.findByEmail(email);
+    if (user) {
+      await this.usersService.updateLastLoginAt(user);
+    } else {
+      throw new UnauthorizedException({ error: 'user_invalid' });
```
ygao76, line 87: We need to use redirect to throw error because this will be user facing. 
Im thinking redirect to https://testnet.ironfish.network/login?toast=encrypted error message

src/auth/auth.controller.ts
```diff
@@ -54,4 +61,78 @@ export class AuthController {
...
+    };
+
+    if (!this.config.isLocal()) {
+      cookieOptions.domain = this.config.get<string>(
+        'INCENTIVIZED_TESTNET_URL',
+      );
+      cookieOptions.sameSite = 'none';
+      cookieOptions.secure = true;
+    }
+
```
ygao76, line 135: Do you want to add the CSRF token here too in the header

src/auth/auth.controller.spec.ts
```diff
@@ -74,7 +78,67 @@ describe('AuthController', () => {
       });
```
ygao76, line 78: nit, would be nice to add the same unit test for jwt token

rohanjadvani: Added

src/auth/auth.controller.ts
rohanjadvani: I was going to do it in a subsequent PR to keep it separate.

src/auth/auth.controller.ts
rohanjadvani: Good call, updated

src/auth/auth.controller.ts
```diff
@@ -54,4 +61,90 @@ export class AuthController {
...
+      }),
+    )
+    { token }: LoginQueryDto,
+    @Res() res: Response,
+  ): Promise<void> {
+    const redirectUrl = `${this.config.get<string>(
+      'INCENTIVIZED_TESTNET_URL',
+    )}/leaderboard`;
+    if (this.config.get<boolean>('DISABLE_LOGIN')) {
+      res.redirect(`${redirectUrl}?toast=${btoa('Login is disabled')}`);
```
ygao76, line 81: Can you add &persist=true so the error message will stay? Current default behavior is disappear in like 2 second, thats too short for user to take screenshot and share us the error message if they have trouble logging in.

src/auth/auth.controller.ts
```diff
@@ -54,4 +61,90 @@ export class AuthController {
...
+    @Query(
+      new ValidationPipe({
+        errorHttpStatusCode: HttpStatus.UNPROCESSABLE_ENTITY,
+        transform: true,
+      }),
+    )
+    { token }: LoginQueryDto,
+    @Res() res: Response,
+  ): Promise<void> {
+    const redirectUrl = `${this.config.get<string>(
```
ygao76, line 77: what do you think about redirect to /login ? no strong opinion but it seems a more common to do that for login error.

ygao76, line 77: Can you define this static constant and export so it can be reused in the spec test?

ygao76 (approved):

PR #1400, opened 2023-04-03
feat: add airdrop fee for miners so airdrop transactions will picked

jowparks:
## Summary
We previously had no fee, this adds a `10 ORE` fee for each airdrop. This will allow us to provide ORE for fee to miners for mining blocks with these transactions.
## Testing Plan
Tests pass

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/jumio-kyc/kyc.service.ts
+++ b/src/jumio-kyc/kyc.service.ts
@@ -13,6 +13,7 @@ import crypto from 'crypto';
 import { ApiConfigService } from '../api-config/api-config.service';
 import {
   AIRDROP_CONFIG,
+  AIRDROP_ORE_FEE,
   ORE_TO_IRON,
   POOL1_TOKENS,
   POOL2_TOKENS,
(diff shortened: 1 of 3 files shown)


src/jumio-kyc/kyc.service.ts
```diff
@@ -362,7 +363,8 @@ export class KycService {
         continue;
       }
 
-      const allocatedOre = Math.floor((userPoolPoints / totalPoints) * poolOre);
+      const allocatedOre =
+        Math.floor((userPoolPoints / totalPoints) * poolOre) - AIRDROP_ORE_FEE;
```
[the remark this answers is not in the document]
jowparks: Yes seems possible, I will just set to 0 if this is the case.

src/jumio-kyc/kyc.service.ts
```diff
@@ -363,7 +364,9 @@ export class KycService {
       }
 
       const allocatedOre = Math.floor((userPoolPoints / totalPoints) * poolOre);
-      await this.redemptionService.update(redemption, { [pool]: allocatedOre });
+      await this.redemptionService.update(redemption, {
```
danield9tqh, line 367: Do we need to use BigInt for all these?

danield9tqh: Max JS number is 2**53 - 1 and the amount of ORE in the genesis block is 42m * ORE_TO_IRON which is ~2**52 so seems safe that none of these numbers will reach an overflow

danield9tqh (approved):

PR #1432, opened 2023-04-19
feat(api): Add nullable `work` to `blocks`

rohanjadvani:
## Summary

Add a column to persist work for blocks

## Testing Plan

N/A

## Breaking Change

```
[ ] Yes
```


jowparks (approved):
I am guessing you will populate this in a subsequent PR? Seems fine

PR #1441, opened 2023-04-20
feat(api): Return block reward

rohanjadvani:
## Summary

Return latest block reward in API response

## Testing Plan

Unit test

## Breaking Change

```
[ ] Yes
```


src/blocks/interfaces/serialized-block-head.ts
```diff
@@ -5,4 +5,5 @@ import { SerializedBlock } from './serialized-block';
 
 export interface SerializedBlockHead extends SerializedBlock {
   hash_rate: number;
+  reward: string;
```
ygao76, line 8: I am okay with the reward in the string since the unit is ore. But curious about how we will use this reward attribute in the charts?

ygao76 (approved):
LGTM, left one comment

PR #1443, opened 2023-04-21
fix(api): Update average difficulty to decimal type

rohanjadvani:
## Summary

BigInts in Prisma don't work if they don't fit into 64 bit integers

## Testing Plan

Manually tested

## Breaking Change

```
[ ] Yes
```


EvanJRichard (approved):
LGTM! :shipit:

PR #1444, opened 2023-04-21
fix(api): Use Decimal for difficulty and work

rohanjadvani:
## Breaking Change

```
[ ] Yes
```


prisma/schema.prisma
```diff
@@ -134,8 +134,8 @@ model Block {
   time_since_last_block_ms Int?
   graffiti                 String             @db.VarChar
   size                     Int?
-  difficulty               BigInt
-  work                     BigInt?
+  difficulty               Decimal?
```
ygao76, line 137: I dont have strong preference on this, but it seems the difficulty type could be `Decimal` instead of nullable Decimal.

ygao76 (approved):
Looks good, left one comment

PR #1462, opened 2023-05-01
feat(ifl-789): verified asset flag

jowparks:
## Summary
Adds verified asset flag to `Asset` model, also returns `verified` data in serialized model.

Plan to run:
```
INSERT INTO assets (name, owner, identifier, metadata, supply, created_transaction_id, verified_at) VALUES ('$IRON', '0000000000000000000000000000000000000000000000000000000000000000', '51f33a2f14f92735e562dc658a5639279ddca3d5079a6d1242b2a588a9cbf44c', 'Iron Fish Native Asset', 0, 56003467, '2023-04-20 09:00:00 PST');
```

Open questions:
1. What should we put for owner?
2. Should we insert a row for `AssetDescriptions`?
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -355,6 +355,7 @@ model Asset {
   name                   String
   owner                  String
   supply                 BigInt
+  verified_at            DateTime?          @db.Timestamp(6)
   descriptions           AssetDescription[]
   created_transaction    Transaction        @relation(fields: [created_transaction_id], references: [id])
 
--- a/prisma/migrations/20230502165227_add_verified_at_asset/migration.sql
+++ b/prisma/migrations/20230502165227_add_verified_at_asset/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "assets" ADD COLUMN     "verified_at" TIMESTAMP(6);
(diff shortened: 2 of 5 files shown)


prisma/schema.prisma
```diff
@@ -355,6 +355,7 @@ model Asset {
   name                   String
   owner                  String
   supply                 BigInt
+  verified               Boolean            @default(false)
```
rohanjadvani, line 358: nit: This is personal preference, but I opt for nullable timestamps instead of booleans if it makes sense (i.e. using `verified_at` instead of `verified`). If there's ever a need to see when an asset was verified, we'll have it and `verified` can be inferred if it's not `NULL`.

rohanjadvani (approved):

rohanjadvani:
* We can create a 0 string for the `owner`
* 42M is inaccurate for the supply since miner rewards will bring more tokens into circulation
* nit: `Ironfish` -> `Iron Fish`
* No, this asset is not created from a mint description

prisma/schema.prisma
jowparks: Sgtm

rohanjadvani (approved):

PR #1465, opened 2023-05-04
feat(api): Update supply for native asset

rohanjadvani:
## Summary

Update the native asset supply every time blocks have been synced

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```


src/blocks/blocks.service.ts
```diff
@@ -156,6 +157,56 @@ export class BlocksService {
...
+
+    // Subject to 1y lock and 1y unlock:
+    // 05.10% - Preseed
+    // 09.90% - Seed
+    // 14.50% - Series A
+    // 00.60% - Advisors
+    // 37.40% - Team
+    // 05.00% - Future Endowments
+    const yearLockAndYearUnlock = 0.725;
+    if (monthsAfterLaunch >= 12) {
```
jowparks, line 193: Just to verify, this means half will be unlocked at 12 months, then the rest is slowly unlocked each month for the next year.

rohanjadvani: Yeah

jowparks (approved):

PR #1471, opened 2023-05-05
feat(api): Remove recaptcha module

rohanjadvani:
## Summary

Delete recaptcha module

## Testing Plan

Existing tests

## Breaking Change

```
[ ] Yes
```


jowparks (approved):
I will leave it up to you if you want to delete this. Seems like it could potentially be useful to have if we ever need to do something with the testnet website in the future.

rohanjadvani:
We're removing user registration and users from our mainnet db. We can dig this up from commit history if needed; I'd rather the API code only contain what our current use case is.

PR #1508, opened 2023-05-25
Whitelist chain_databaseSize in telemetry

ygao76:
## Summary
Whitelist chain_databaseSize in telemetry service
## Breaking Change

```
[ ] Yes
```
@rohanjadvani @danield9tqh @dguenther


src/telemetry/telemetry.controller.ts
```diff
@@ -80,6 +80,7 @@ const TELEMETRY_WHITELIST = new Map<string, true | Array<string>>([
   ],
   ['transaction_propagation', true],
   ['forks_count', true],
+  ['chain_databaseSize', true],
```
danield9tqh, line 83: `chain_database_size`?

danield9tqh (approved):

PR #1525, opened 2023-06-07
feat(api): Add supply controller

rohanjadvani:
## Summary

Adds separate endpoints for fetching circulating and total supplies

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/supply/supply.controller.ts
+++ b/src/supply/supply.controller.ts
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Controller, Get } from '@nestjs/common';
+import { ApiOperation, ApiTags } from '@nestjs/swagger';
+import { BlocksService } from '../blocks/blocks.service';
+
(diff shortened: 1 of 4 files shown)


src/supply/supply.controller.ts
```diff
@@ -0,0 +1,35 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Controller, Get } from '@nestjs/common';
+import { ApiOperation, ApiTags } from '@nestjs/swagger';
+import { BlocksService } from '../blocks/blocks.service';
+
+@ApiTags('Supply')
+@Controller('supply')
```
ygao76, line 12: The url `supply/total` and `supply/circulating` has the revert order of total supply and circulating supply, do we want the url name be similar to solana or avascan? like `total-supply` or `supply?q=totalSupply`
http://api.mainnet-beta.solana.com/v0/circulating-supply
https://avascan.info/api/v1/supply?q=totalSupply

NullSoldier: I think that's a reasonable comment, but in this case the URL's are addressing hierarchical resources which `circulating` and `total` are sub resources in the `supply` "folder". I think it's ok that it "seems" reversed but in this case were just navigating down the hierarchy to sub-resources.

NullSoldier (approved):
This looks good. I don't personally have an issue with the url convention, but if we want to discuss that further we can come to an agreement.

src/supply/supply.controller.ts
ygao76: sounds good

PR #1528, opened 2023-06-08
Whitelist chaindb in telemetry service in api

ygao76:
## Summary
Whitelist chaindb in telemetry service in api https://github.com/iron-fish/ironfish/pull/3954

## Breaking Change

```
[ ] Yes
```
@dguenther @iron-fish/engineering


src/telemetry/telemetry.controller.ts
```diff
@@ -80,7 +80,7 @@ const TELEMETRY_WHITELIST = new Map<string, true | Array<string>>([
   ],
   ['transaction_propagation', true],
   ['forks_count', true],
```
jowparks, line 82: Was this intentional removing the `chain_database_size`?

ygao76: yes, removing it and created a new metric. The new metric is submitted from stats cmd from api-stats instance. The old one is in sdk and we decide its better to use stats cmd.

dguenther (approved):

PR #1535, opened 2023-06-15
Add `INCENTIVIZED_TESTNET_URL` to `.env.template`

andreacorbellini:
## Summary

`INCENTIVIZED_TESTNET_URL` is a required configuration variable; without it, the server won't start.

## Testing Plan

Check out a clean copy of the repository (or delete your `.env` file) and follow the instructions in the README: without this change, the server won't start; after this change, the server starts.

## Breaking Change

Not a breaking change.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,5 +1,6 @@
 API_URL=http://localhost:8003
 BLOCK_EXPLORER_URL=http://localhost:3000
+INCENTIVIZED_TESTNET_URL=http://localhost:3001
 BLOCK_LOADER_TRANSACTION_TIMEOUT=5000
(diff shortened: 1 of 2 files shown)


.env.template
```diff
@@ -1,5 +1,6 @@
 API_URL=http://localhost:8003
 BLOCK_EXPLORER_URL=http://localhost:3000
+INCENTIVIZED_TESTNET_URL=http://localhost:3001
```
ygao76, line 3: Is the testnet url still in use? It might be only for old snapshot data

ygao76: Yeah if its not used anywhere, let's remove the variable.

andreacorbellini: https://github.com/iron-fish/ironfish-api/pull/1536

PR #1563, opened 2023-06-29
feat(api): Remove read pool URL for database

rohanjadvani:
## Summary

We don't need to use the read replica, so remove references.

## Testing Plan

N/A

## Breaking Change

```
[ ] Yes
```


andreacorbellini:
If one day we wanted to scale up our database and move it to a primary/replica setup, wouldn't we need this code?

PR #1572, opened 2023-07-07
Add the `/assets/verified` endpoint

andreacorbellini:
## Summary

This endpoint is more lightweight than `/assets?verified=true` and differs from it in that:

- it returns only asset identifiers, no other metadata;
- it does not support search filters;
- it's not paginated;
- the object returned by the new endpoint uses a more semantical field (`assets` instead of `data`).

## Testing Plan

```
$ curl -sv http://localhost:8003/assets/verified | jq
...
< HTTP/1.1 200 OK
< Last-Modified: Wed, 21 Jun 2023 17:58:10 GMT
...
{
  "assets": [
    {
      "identifier": "51f33a2f14f92735e562dc658a5639279ddca3d5079a6d1242b2a588a9cbf44c"
    }
  ]
}
```

## Breaking Change

Not a breaking change


src/assets/assets.service.ts
```diff
@@ -165,6 +166,28 @@ export class AssetsService {
     };
   }
 
+  listIdentifiers(options: ListAssetIdsOptions): Promise<Partial<Asset>[]> {
```
rohanjadvani, line 169:
```suggestion
  async listIdentifiers(options: ListAssetIdsOptions): Promise<Partial<Asset>[]> {
```

src/assets/assets.controller.ts
```diff
@@ -128,4 +128,18 @@ export class AssetsController {
...
+  @Header('Cache-Control', 's-maxage=60, stale-if-error=60')
+  async verified(@Req() req: Request, @Res() res: Response): Promise<void> {
+    const lastUpdate = await this.assetsService.lastUpdate();
+    if (lastUpdate) {
+      handleIfModifiedSince(lastUpdate, req, res);
+    }
+
+    const assets = await this.assetsService.listIdentifiers({ verified: true });
+
+    res.json({ assets: assets });
```
rohanjadvani, line 143:
```suggestion
    res.json({ assets });
```

rohanjadvani (approved):
Looks great, some minor TS nits

src/assets/assets.service.ts
andreacorbellini: This is returning a `Promise` directly without using any `await` expression; it shouldn't be marked `async`

PR #1654, opened 2023-10-04
Rahul/ifl 1733 add an endpoint to update block graffiti in syncer

patnir:
## Summary

Add update block endpoint

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/blocks/blocks.controller.spec.ts
+++ b/src/blocks/blocks.controller.spec.ts
@@ -215,6 +215,40 @@ describe('BlocksController', () => {
         .expect(HttpStatus.UNAUTHORIZED);
     });
 
+    it('throw error when graffiti is not formatted correctly', async () => {
+      // graffiti is not hex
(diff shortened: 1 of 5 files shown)


src/blocks/blocks.controller.spec.ts
```diff
@@ -208,6 +208,58 @@ describe('BlocksController', () => {
...
+        );
+
+      await request(app.getHttpServer())
+        .post('/blocks/update_graffiti')
+        .set('Authorization', `Bearer ${API_KEY}`)
+        .send({ hash: 'hash', graffiti: 'graffiti' })
+        .expect(HttpStatus.CREATED);
+
+      expect(updateGraffiti).toHaveBeenCalledWith(
+        {
```
danield9tqh, line 254: Does this function expect an object?

danield9tqh (approved):

PR #1655, opened 2023-10-05
batch update graffiti api

patnir:
## Summary

At the moment it takes 90 minutes to update 25k records using the update_graffiti API

Batch update graffiti for performance improvements over single update

## Breaking Change

```
[ ] Yes
```


src/blocks/blocks.service.ts
```diff
@@ -117,6 +118,50 @@ export class BlocksService {
...
+    });
+
+    blocks.map((block) => {
+      const graffiti = hashGraffitiMap.get(block.hash);
+      if (graffiti) {
+        block.graffiti = graffiti;
+      }
+    });
+
+    await this.prisma.$transaction([
```
danield9tqh, line 149: One thing I'd be careful about here is putting too many updates into a single transactions. Transactions that are too big can really stall out the database just FYI

danield9tqh (approved):

PR #1700, opened 2024-01-11
remove blocks/status endpoint

NullSoldier:
## Summary

No idea what this endpoint is, but I cannot find any place it's used. Something is hitting it in the API and causing it to crash because of a transaction timeout exceeded in a loop.

It returns

```
{"object":"blocks_status","chain_height":385426,"percentage_marked":1,"unique_graffiti":385423}
```

It was added here, https://github.com/iron-fish/ironfish-api/pull/337
Here is the original ticket https://linear.app/ironfish/issue/IRO-1210/add-graffiti-stats-endpoint-to-api

## Breaking Change

```
[X] Yes
```


jowparks:
I did a quick look around and didn't see it in the cli or the api code, so this seems ok to me. It is a bit worrying that something is hitting the API, I would have suspected it was us so I am hoping we aren't breaking something with this.

jowparks (approved):

PR #1716, opened 2024-03-20
Add verified assets metadata

danield9tqh:
## Summary
Adding verified asset metadata with multiple fields to assets. Adding endpoint to sync verified assets from a external JSON list of verified assets

## Testing Plan
Unit tests and testing endpoints locally

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/src/assets/assets.controller.spec.ts
+++ b/src/assets/assets.controller.spec.ts
@@ -12,7 +12,10 @@ import { PrismaService } from '../prisma/prisma.service';
 import { bootstrapTestApp } from '../test/test-app';
 import { TransactionsService } from '../transactions/transactions.service';
 import { AssetsService } from './assets.service';
+import { VerifiedAssetMetadataDto } from './dto/update-verified-assets-dto';
 import { Asset, Block, Transaction } from '.prisma/client';
+
+const API_KEY = 'test';
 
 describe('AssetsController', () => {
   let app: INestApplication;
@@ -44,6 +47,7 @@ describe('AssetsController', () => {
     block: Block;
     transaction: Transaction;
     asset: Asset;
+    metadata: VerifiedAssetMetadataDto;
   }> {
     const block = await blocksService.upsert(prisma, {
       hash: uuid(),
@@ -82,7 +86,15 @@ describe('AssetsController', () => {
       prisma,
     );
 
-    return { block: block, transaction: transaction, asset: asset };
+    const metadata = {
+      identifier: asset.identifier,
+      symbol: uuid(),
+      decimals: Math.floor(Math.random() * 10),
+      logoURI: uuid(),
+      website: uuid(),
+    };
+
+    return { block: block, transaction: transaction, asset: asset, metadata };
   }
 
   async function verifyAsset(asset: Asset): Promise<Asset> {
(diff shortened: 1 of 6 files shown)


src/assets/assets.controller.ts
```diff
@@ -142,4 +150,30 @@ export class AssetsController {
 
     res.json({ assets });
   }
+
+  @ApiExcludeEndpoint()
+  @Post('update-verified')
```
mat-if, line 155: What do you think about making this `@Post('verified')`? It's more inline with RESTful API guidelines

src/assets/assets.controller.ts
```diff
@@ -142,4 +150,30 @@ export class AssetsController {
...
+        errorHttpStatusCode: HttpStatus.UNPROCESSABLE_ENTITY,
+        transform: true,
+      }),
+    )
+    dto: UpdateVerifiedAssetsDto,
+    @Res() res: Response,
+  ): Promise<void> {
+    const missing: VerifiedAssetMetadataDto[] = [];
+    for (const asset of dto.assets) {
+      const result = await this.assetsService.updateVerified(asset);
```
mat-if, line 169: Should `updateVerified` and `deleteVerified` be in a transaction together? That way if anything breaks, we get a clean atomic rollback

src/assets/assets.controller.ts
```diff
@@ -142,4 +150,30 @@ export class AssetsController {
 
     res.json({ assets });
   }
+
+  @ApiExcludeEndpoint()
+  @Post('update-verified')
+  @UseGuards(ApiKeyGuard)
+  async update_verified(
```
mat-if, line 157:
```suggestion
  async updateVerified(
```

src/assets/assets.controller.spec.ts
```diff
@@ -556,4 +568,193 @@ describe('AssetsController', () => {
       });
     });
   });
+
+  describe('POST /assets/update_verified', () => {
+    it('Adds metadata for all assets that exist', async () => {
```
mat-if, line 573:
```suggestion
    it('adds metadata for all assets that exist', async () => {
```

src/assets/assets.controller.spec.ts
```diff
@@ -556,4 +568,193 @@ describe('AssetsController', () => {
...
+          decimals: metadata.decimals,
+          logo_uri: metadata.logoURI,
+          website: metadata.website,
+          created_at: expect.any(Date),
+          updated_at: expect.any(Date),
+        });
+      }
+    });
+
+    it('Updates metadata for assets that already have metadata', async () => {
```
mat-if, line 601:
```suggestion
    it('updates metadata for assets that already have metadata', async () => {
```

src/assets/assets.controller.spec.ts
```diff
@@ -556,4 +568,193 @@ describe('AssetsController', () => {
...
+          decimals: assetMetadata2.decimals,
+          logo_uri: assetMetadata2.logoURI,
+          website: assetMetadata2.website,
+          created_at: expect.any(Date),
+          updated_at: expect.any(Date),
+        },
+      ]);
+    });
+
+    it('Only updates metadata for assets that exist and returns those not found', async () => {
```
mat-if, line 662:
```suggestion
    it('only updates metadata for assets that exist and returns those not found', async () => {
```

src/assets/assets.controller.spec.ts
```diff
@@ -556,4 +568,193 @@ describe('AssetsController', () => {
...
+          decimals: assets[0].decimals,
+          logo_uri: assets[0].logoURI,
+          website: assets[0].website,
+          created_at: expect.any(Date),
+          updated_at: expect.any(Date),
+        },
+      ]);
+    });
+
+    it('Deletes metadata items that are not in the list', async () => {
```
mat-if, line 708:
```suggestion
    it('deletes metadata items that are not in the list', async () => {
```

src/assets/assets.controller.spec.ts
```diff
@@ -556,4 +568,193 @@ describe('AssetsController', () => {
...
+
+      await expect(prisma.verifiedAssetMetadata.findMany({})).resolves.toEqual(
+        [],
+      );
+
+      // Make sure asset itself is not deleted
+      await expect(prisma.asset.findMany({})).resolves.toEqual([asset]);
+    });
+
+    it('Deletes metadata if the asset is deleted', async () => {
```
mat-if, line 741:
```suggestion
    it('deletes metadata if the asset is deleted', async () => {
```

src/assets/assets.controller.ts
danield9tqh: `@Post('verify')`?

src/assets/assets.controller.ts
danield9tqh: yea thought about this. my only concern was that the transaction would be too big and interrupt other API stuff with a big SQL job. I think if a partial update went through it wouldn't be the worst thing in the world. Its idempotent after all so it could just be run again. Wdyt?

mat-if (approved):

PR #1718, opened 2024-03-26
Upgrade from NodeJS 18 to 20

mat-if:
## Summary

- Changed package.json to use >=20 for NodeJS version so hopefully it will be easier to use 22+ down the line
- Upgraded the Github Actions to use NodeJS 20
- Ran `yarn upgrade hot-shot` to upgrade a sub-dependency that was failing
- Modified a test to cast to a number. I'm not exactly sure why this started failing now, but it is more correct now, technically.

## Testing Plan

Unit tests locally
Unit tests on github
Deploy to testnet

## Breaking Change

Only for development if anyone was still using NodeJS 18, so no not really.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,40 +4,40 @@
 
 ## Installing
 
-* Make sure you're running at least Node 16 - you may want to avail yourself of a tool like [nvm](https://nvm.sh)
-* Run `yarn` to install dependencies
-* Run `yarn docker:start` to start docker
-* Run `yarn build` to build things locally
-* Run `yarn db:client:generate` to generate the Prisma client
-* Run `yarn db:migrate` to create a migration
-* (Optional) Run `yarn db:seed:testing` to load some sample data into the database
+- Make sure you're running at least Node 20 - you may want to avail yourself of a tool like [nvm](https://nvm.sh)
+- Run `yarn` to install dependencies
+- Run `yarn docker:start` to start docker
+- Run `yarn build` to build things locally
+- Run `yarn db:client:generate` to generate the Prisma client
+- Run `yarn db:migrate` to execute migrations and optionally create a migration if you have made schema changes
+- (Optional) Run `yarn db:seed:testing` to load some sample data into the database
 
 ## Environment
 
 You'll need a `.env` file with the following keys:
 
-* `API_URL`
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -4,40 +4,40 @@
...
 ## Installing
 
-* Make sure you're running at least Node 16 - you may want to avail yourself of a tool like [nvm](https://nvm.sh)
-* Run `yarn` to install dependencies
-* Run `yarn docker:start` to start docker
-* Run `yarn build` to build things locally
-* Run `yarn db:client:generate` to generate the Prisma client
-* Run `yarn db:migrate` to create a migration
-* (Optional) Run `yarn db:seed:testing` to load some sample data into the database
+- Make sure you're running at least Node 20 - you may want to avail yourself of a tool like [nvm](https://nvm.sh)
```
mat-if, line 7: VSCode's auto-formatter wanted to do this, I guess. In Markdown, `*` and `-` _are_ equivalent, so 🤷

src/blocks/blocks.service.ts
```diff
@@ -378,7 +378,7 @@ export class BlocksService {
     }
 
     return {
-      averageBlockTimeMs: dateMetricsResponse[0].average_block_time_ms,
+      averageBlockTimeMs: Number(dateMetricsResponse[0].average_block_time_ms),
```
mat-if, line 381: Not sure why this wasn't needed before, exactly. But the unit test started failing due to this being an object (and it _was_ indeed an object before being cast according to `typeof`)

yarn.lock
```diff
@@ -1639,7 +1639,8 @@ binary-extensions@^2.0.0:
 
 bindings@^1.5.0:
   version "1.5.0"
-  resolved "https://registry.npmjs.org/bindings/-/bindings-1.5.0.tgz"
+  resolved "https://registry.yarnpkg.com/bindings/-/bindings-1.5.0.tgz#10353c9e945334bc0511a6d90b38fbc7c9c504df"
```
mat-if, line 1642: These changes are from running `yarn upgrade hot-shot` to fix an issue with `unix-dgram` sub-dependency.

image

rohanjadvani (approved):

PR #1719, opened 2024-03-26
Add verified asset metadata endpoint

mat-if:
## Summary

Adds the endpoint serving the new data. It's basically just a pass-through of the data in the DB, and isn't much different from just using the JSON file from github CDN or something, but it gives us the ability to control the endpoint.

## Testing Plan

Unit tests, manual testing

## Breaking Change

No


src/assets/assets.controller.ts
```diff
@@ -176,4 +177,34 @@ export class AssetsController {
...
   }
+
+  @ApiOperation({ summary: 'Lists all verified asset metadata' })
+  @Get('verified_metadata')
+  @Header('Cache-Control', 's-maxage=60, stale-if-error=60')
+  async verifiedMetadata(
+    @Req() req: Request,
+    @Res() res: Response,
+  ): Promise<void> {
+    const lastUpdate = await this.assetsService.lastUpdate();
```
danield9tqh, line 188: This last update currently is looking at the asset table but I think we want to look at the assetMetadata table for the new one. I think the easiest thing to do would be to change the old endpoint to return the new data before this PR (new data database must be in mainnet and testnet databases before we do that). Once the old endpoint returns the new data we can re-use this function in the new endpoint

prisma/seed/main.ts
```diff
@@ -112,6 +112,32 @@ async function addAsset({
   console.log(`Created asset ${name}`);
 }
 
+async function addVerifiedAssetMetadata({
```
danield9tqh, line 115: Think I already added this in the previous PR

src/assets/assets.controller.ts
```diff
@@ -176,4 +177,34 @@ export class AssetsController {
...
+    if (lastUpdate) {
+      handleIfModifiedSince(lastUpdate, req, res);
+    }
+
+    const assetsMetadata = (await this.assetsService.listMetadata()).map(
+      (m) => {
+        return {
+          identifier: m.identifier,
+          symbol: m.symbol,
+          decimals: m.decimals,
```
danield9tqh, line 198: Probably should not put fields on here if they are `null` because the spec for verified asset metadata JSON shows these fields as optional but not nullable

danield9tqh (approved):

PR #1767, opened 2024-09-28
Add initial bridge implementation

dguenther:
## Summary

Adds a `/bridges` route with APIs for fetching bridgeable assets, fetching destination networks for those assets, creating transaction outputs, and checking transaction status.

Fixes IFL-2984
Fixes IFL-2986
Fixes IFL-2987
Fixes IFL-2988
Fixes IFL-2994

## Testing Plan

I wrote tests for the ChainportService, but since it hits the live API I've skipped them. I think they're still useful to have committed in case we need to update the implementation. 

## Breaking Change

```
[ ] Yes
```


src/bridges/chainport.service.ts
```diff
@@ -0,0 +1,345 @@
...
+    return firstValueFrom(
+      this.httpService.get<T>(url).pipe(
+        catchError((e: AxiosError) => {
+          this.logger.error(
+            `Chainport Error: ${url} - Status Code ${
+              e.status ?? 'Unknown'
+            } - ${JSON.stringify(e.response?.data)}`,
+            new Error().stack ?? '',
+          );
+          throw e;
```
rohanjadvani, line 146: NestJS has `BadGatewayException` if we want to throw a 502 here. I think if we see that in the responses then we can check the logs accordingly.

src/bridges/chainport.service.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    const validateResult = chainportNetworkArraySchema.validate(
+      Object.values(meta.cp_network_ids),
+      {
+        stripUnknown: true,
+      },
+    );
+
+    if (validateResult.error) {
+      throw new Error(
```
rohanjadvani, line 183: Same comment here for `BadGatewayException`.

src/bridges/chainport.service.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    const validateResult = chainportTokenArraySchema.validate(
+      result.data.verified_tokens,
+      {
+        stripUnknown: true,
+      },
+    );
+
+    if (validateResult.error) {
+      throw new Error(
```
rohanjadvani, line 215: Same comment re `BadGatewayException`

src/bridges/chainport.service.ts
```diff
@@ -0,0 +1,345 @@
...
+      if (!sourceToken) {
+        throw new NotFoundException();
+      }
+
+      networkList = sourceToken.target_networks.flatMap((n) => {
+        const network = metaResult.cp_network_ids[n.toString()];
+        if (!network) {
+          this.logger.error(
+            `Network ${n} for token ${tokenId} not found in meta`,
+            new Error().stack ?? '',
```
rohanjadvani, line 249: Does `new Error().stack` always return some stack we can use? Or should we just use ''`?

src/bridges/chainport.service.ts
```diff
@@ -0,0 +1,345 @@
...
+    url: string,
+  ): Promise<AxiosResponse<T, unknown>> {
+    return firstValueFrom(
+      this.httpService.get<T>(url).pipe(
+        catchError((e: AxiosError) => {
+          this.logger.error(
+            `Chainport Error: ${url} - Status Code ${
+              e.status ?? 'Unknown'
+            } - ${JSON.stringify(e.response?.data)}`,
+            new Error().stack ?? '',
```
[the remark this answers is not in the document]
dguenther: Good idea 👍

src/bridges/chainport.service.ts
dguenther: Yep! It returns the stack at the current location

rohanjadvani (approved):

PR #1805, opened 2025-07-07
feat: support chainport bridge fee v2

iflabsfish:
## Breaking Change

```
[ ] Yes
```


danield9tqh (approved):

danield9tqh:
This assumes a single API switch over from V1 to V2 based on a config variable. Is it possible that we would have old clients that haven't migrated yet and we still need to support V1 and V2 simultaneously in the API? Or are the two fundamentally incompatible?

iflabsfish:
> This assumes a single API switch over from V1 to V2 based on a config variable. Is it possible that we would have old clients that haven't migrated yet and we still need to support V1 and V2 simultaneously in the API? Or are the two fundamentally incompatible?

`ironfish-api` is a wrapper for chainport api service, since this is a breaking  upgrade on their side, it's meaningless to make it compatible on V1 and V2 on our side.

